PR #804, opened 2014-09-23
Cleanup Analytics Module

kabel:
Update the automatic link tracking for the analytics module to match what the UNLcms currently provides.

This will allow developers to control when/which links are automatically tracked using the `WDN.setPluginParam` method. Examples:
To change the file extensions tracked - `WDN.setPluginParam('analytics', 'extensions', 'pdf|zip');`
To disable file download tracking - `WDN.setPluginParam('analytics', 'trackDownload', false);`
To disable mailto link tracking - `WDN.setPluginParam('analytics', 'trackMailto', false);`
To disable outbound link tracking - `WDN.setPluginParam('analytics', 'trackOutbound', false);`

Also disable the scroll depth plugin by default, allowing developers to manually enable it on pages they'd like those events to be tracked. (start discussion here)

### DIFF
--- a/wdn/templates_4.0/scripts/analytics.js
+++ b/wdn/templates_4.0/scripts/analytics.js
@@ -10,56 +10,62 @@ define(['wdn', 'idm', 'jquery'], function(WDN, idm, $) {
 		wdnProp = 'UA-3203435-1',
 		unlDomain = '.unl.edu',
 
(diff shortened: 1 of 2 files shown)


kabel:
I couldn't get any momentum behind this, so we'll wait for other to raise concern.

PR #830, opened 2015-03-09
Fix the algorithm of allowing UNL Search URL extending

kabel:
Local sites should only be allowed to append the U or CX querystring
params for use in the WDN search tool.

Logic has been added to enforce the use of ONLY the UNL Search app from
the WDN search tool.

This commit also removes the experimental speak-to-text attribute that
has been removed from the only browser that supported it.

The undocumented feature to add a local placeholder has also been
removed (unapproved by the style guide)

### DIFF
--- a/wdn/templates_4.0/scripts/search.js
+++ b/wdn/templates_4.0/scripts/search.js
@@ -23,117 +23,145 @@ define(['jquery', 'wdn', 'require', 'modernizr', 'navigation'], function($, WDN,
 					submitted = false,
 					postReady = false,
 					autoSubmitTimeout,
-					searchOrigin = '//www1.unl.edu',
-					siteHomepage = nav.getSiteHomepage();
+					searchHost = 'www1.unl.edu', // domain of UNL Search app
+					searchPath = '/search/', // path to UNL Search app
+					searchOrigin = window.location.protocol + '//' + searchHost,
+					searchAction = searchOrigin + searchPath,
+					allowSearchParams = ['u', 'cx'],  // QS Params allowed by UNL Search app
+					siteHomepage = nav.getSiteHomepage(),
+					localSearch = getLocalSearch();
 
-				/**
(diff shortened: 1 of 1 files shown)


wdn/templates_4.0/scripts/search.js
```diff
@@ -23,117 +23,145 @@ define(['jquery', 'wdn', 'require', 'modernizr', 'navigation'], function($, WDN,
...
-							htmlUpdate = htmlUpdate.add($('<input>', {
-								type: "hidden",
-								name: hash[0],
-								value: decodeURIComponent(hash[1])
-							}));
+					searchHost = 'www1.unl.edu', // domain of UNL Search app
+					searchPath = '/search/', // path to UNL Search app
+					searchOrigin = window.location.protocol + '//' + searchHost,
+					searchAction = searchOrigin + searchPath,
+					allowSearchParams = ['u', 'cx'],  // QS Params allowed by UNL Search app
```
mfairchild365: Out of curiosity, what is `u`?  I didn't see it on this list: https://developers.google.com/custom-search/json-api/v1/reference/cse/list

kabel: `u` is a custom parameter the UNL Search app uses to auto-magically create a Google CSE on-the-fly.

mfairchild365: Ah okay.  Why would a site want to set it, and what would they set it to?

kabel: Too allow developers to manually provide a "search root" for their site/sub-site in cases where the default logic (using the site homepage) would result in unexpected/poor local search performance. For example, when the home page of a site is at a file URL rather than a directory.

This is just a stricter enforcement of the previously documented was of defining your local search: http://www1.unl.edu/wdn/wiki/Local_Site_Links#search

PR #832, opened 2015-04-16
Email missing link fix

spennythug:


kabel:
This should address #833

wdn/templates_4.0/scripts/socialmediashare.js
```diff
@@ -95,5 +146,31 @@ define(['jquery', 'wdn', 'require'], function($, WDN, require) {
         }
     };
 
+    assembleLink = function(shareId, subject, url, isShort) {
```
kabel: I would put these functions at the top of the file _and_ make sure you add the `var ` prefix to ensure they don't get pushed into the `window` global scope.

PR #834, opened 2015-04-23
Fix issue with mediaelement loading image in firefox + update

mfairchild365:
It appears that firefox does not always send the `canplay` event, which causes the loading image to always display.  This is similar to another android bug which was already 'fixed'.  I don't like this method of fixing this issue, but I don't know a better way.

This PR also updates mediaelement


mfairchild365:
@kabel I'm not sure about the correct way to update this plugin as the file structure is not 1:1 with the medialement repo.  I cherry picked the files that I thought needed to be updated, but I didn't check the css/images.  Should I update those too?  If so, which ones?

wdn/templates_4.0/scripts/plugins/mediaelement/mediaelement-and-player.js
```diff
@@ -2957,10 +3031,23 @@ if (typeof jQuery != 'undefined') {
 
 				loading.show();
 				controls.find('.mejs-time-buffering').show();
+                // Firing the 'canplay' event after a timeout which isn't getting fired on some Android 4.1 devices (https://github.com/johndyer/mediaelement/issues/1305)
+                if (mejs.MediaFeatures.isAndroid | mejs.MediaFeatures.isFirefox) {
```
kabel: This should be a logical OR `||` operator, not a bitwise OR `|` operator.

mfairchild365: fixed in https://github.com/unl/wdntemplates/commit/2e387e0ccf7a124a25d00212db5e37d2abb50f7b

mfairchild365:
@kabel I updated the rest of the assets.  Ready for another review.

kabel:
I think we should fix the permissions that were changed. Images and CSS should not have the Execute unix permission on.

mfairchild365:
@kabel done.

kabel:
I found a _hopefully_ better way to handle this by creating a new patch for the MediaElementJS upstream. See f9da1c43ad6efea04f09759c76c30d84e4ae87ab.

The problem with the upstream is that (after reading the spec at http://dev.w3.org/html5/spec-preview/media-elements.html#event-media-loadeddata) the events they add listeners for are not exactly right. The `loadeddata` event should signal the end of loading, but that is used by mejs to show the loading interface.

mfairchild365:
@kabel Nice find! It looks good to me.  Where did you find that upstream patch? I can't find it anywhere.

PR #835, opened 2015-04-27
increase swipe length required to open navigation

mfairchild365:
When scrolling down a long page of information on a touch enabled mobile device, it was easy to accidentally open the side navigation.  This was frustrating and not expected.  This commit aims to address that issue by doubling the distance required to swipe for the navigation to open.  While it might not be a perfect solution, it is hopefully better than it used to be.

Maybe it would be better to remove this feature altogether?


kabel:
Let's accept this for now and propose to remove it in the next Gov Board meeting. I think this has been a failed experiment in used expectation.

PR #837, opened 2015-05-05
Fix events RSS link on upcoming list.

tlemburg:
The link was not going to the correct feed. This two character change makes it have the right format.


tlemburg:
w00t

PR #870, opened 2015-07-23
Navigation style and experience improvements

kabel:
Fixes the following navigation issues:

* Content jumps in mobile when the navigation bar floats
* Swipe to open nav has not worked out (opens at the wrong time)
* Poor implementations of throttle and debounce for DOM events.


kabel:
:+1: or :-1:

wdn/templates_4.0/scripts/navigation.js
```diff
@@ -44,6 +45,72 @@ define(['jquery', 'wdn', 'modernizr', 'require'], function($, WDN, Modernizr, re
...
+	var _now = Date.now || function() {
+		return new Date().getTime();
+	};
+
+	var throttle = function(func, wait, options) {
+		var context, args, result;
+		var timeout = null;
+		var previous = 0;
+		if (!options) options = {};
+		var later = function() {
```
mfairchild365: This looks good, but I have two thoughts.  These functions are pretty hard to follow, especially if you are new to them and what they are supposed to do.  I think some comments could really add to the readability.  For example, what is the purpose of `previous` and what sort of values should it contain?  what is `later` used for?

I could also see an unit tests be very helpful for functions like these.

mfairchild365:
:+1: 

I ran it locally and it appeared to work beautifully.

wdn/templates_4.0/scripts/navigation.js
kabel: Good point. These are pulled directly from vendor code. I will add a comment to that extent.

PR #871, opened 2015-08-11
Extend font features (small caps, extended Latin)

skoolbus39:
Extended font features can optionally be included via JavaScript.

### DIFF
--- a/wdn/templates_4.0/scripts/extlatin.js
+++ b/wdn/templates_4.0/scripts/extlatin.js
@@ -1,0 +1,14 @@
+define(['jquery', 'wdn', 'require'], function($, WDN, require) {
+
+    // Load Mercury ScreenSmart Extended Latin fonts from Cloud.typography
+    var $extlatin = $("<link>", {
+        "rel" : "stylesheet",
+        "type" : "text/css",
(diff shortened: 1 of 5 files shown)


wdn/templates_4.0/scripts/extlatin.js
```diff
@@ -0,0 +1,14 @@
+define(['jquery', 'wdn', 'require'], function($, WDN, require) {
+
+    // Load Mercury ScreenSmart Extended Latin fonts from Cloud.typography
```
kabel: I'd wrap all of this in the following to make it compatible with the `WDN.initializePlugin` signature:

```
var initd = false;

return {
    initializePlugin: function() {
        // protect against multiple initializations
        if (initd) {
            return;
        }
        initd = true;

        [...]
    }
}
```

wdn/templates_4.0/scripts/smallcaps.js
```diff
@@ -0,0 +1,14 @@
+define(['jquery', 'wdn', 'require'], function($, WDN, require) {
+
+    // Load Mercury ScreenSmart Small Caps fonts from Cloud.typography
```
kabel: Same comment from about about compatibility.

PR #877, opened 2015-08-19
Fix the Windows XP notice display

kabel:
* The campus service has moved and now works on https
* XP can present itself at Windows NT 5.2 (from browserstack tests)


kabel:
Care to take a look at this one too @mfairchild365?

PR #883, opened 2015-08-28
Use only nickname or first name in IdM module

### LINKED ISSUE #751: Improper display of user's name (cultural issue)
The IDM area is now displaying the logged in user's name in the format "Eric R."

I believe with most Asian names, such as Yao Ming, you would only ever display them as "Yao" or "Yao Ming" but never "Yao M." or "Ming".

So I suppose the suggested fix is to only show first name.

Related to change e4784d239b6956a9b5850b6e622b1f43740f5e24

kabel:
If these fields are available the first letter of the surname will no
longer be added. Otherwise, the full "display name" will be used.
Closes #751.


saltybeagle:
:+1:

PR #884, opened 2015-08-28
Give #maincontent focus on skipnav click

### LINKED ISSUE #829: skipnav doesn't work in some browsers
I've been experimenting with VoiceOver lately, and have noticed that in Chome and Firefox when you click the `skip navigation` link, it visually positions you at the `#maincontent` area, but when you press tab or try to interact with the page, it brings you to the next element after the `skip navigation` link, which seems frustrating and unexpected.

I've done some expirmenting and a little research and it appears that the solution is to add a `tabindex="-1"` to the maincontent area.

https://carmenwiki.osu.edu/display/10292/HTML5+Accessibility+Code+Examples
http://www.bignerdranch.com/blog/web-accessibility-skip-navigation-links/

I think this would require editing the Template HTML, which we can't do until the next major release.  One solution might be to do it with JavaScript.

mfairchild365:
This fixes an issue where if the skipnav link was clicked, focus was not being set to the #maincontent div.  The screen would just visually focus on the area, rather than move tab focus to the maincontent area.

fixes #829

### DIFF
--- a/wdn/templates_4.0/scripts/main.js
+++ b/wdn/templates_4.0/scripts/main.js
@@ -18,13 +18,14 @@ require.config({
 //Modernizr, WDN are loaded prior to requireJS
 define('modernizr', [], function () { return window.Modernizr; });
 
-define(['wdn', 'skipnav', 'require', 'legacy',
+define(['wdn', 'require', 'legacy',
         // these are the WDN plugins that are required
         'analytics',
         'navigation',
         'search',
         'unlalert',
-        'images'], function(WDN, require) {
+        'images', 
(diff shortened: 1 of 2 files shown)


kabel:
This is good for now. I'd love to see something more like what was done in Sheldon, by removing the attribute on blur.

kabel:
Problem: You changed the order of the main dependencies, and thus the callback parameters don't match.

mfairchild365:
@kabel can you link me to what was done with Sheldon, I can't find it.

kabel:
Private repo link shared out-of-band.

PR #886, opened 2015-09-02
Improve a11y for the search feature

### LINKED ISSUE #885: Keyboard navigation, a11y, closing search results
Currently there's no easy way to close the popup search results box using the keyboard.

Would it be possible to allow the user to remove text input in the search box, or hit escape to exit/close out of the popup search results box?

mfairchild365:
This does two things to improve a11y for the search feature.

1. fixes #885 - Pressing escape will close the results pane.  This could be further improved by adding a link to close the results pane (but I'm not sure exactly how the would look).
2. Send focus to the results pane on manual submit (not autocomplete)

Note that an update to the search application will have to be rolled out at the same time.  There is a paired PR for that project which I will submit in just a second.

### DIFF
--- a/wdn/templates_4.0/scripts/search.js
+++ b/wdn/templates_4.0/scripts/search.js
@@ -155,11 +155,41 @@ define(['jquery', 'wdn', 'require', 'modernizr', 'navigation'], function($, WDN,
 					}
 				});
 
+				var closeSearch = function() {
+					var $wdnSearch = domSearchForm.parent();
+					$wdnSearch.removeClass('active');
+				};
+
+				//Close search on escape while the iframe has focus
+				$(window).on('message', function(e) {
+					var originalEvent = e.originalEvent;
+					
+					if ('wdn.search.close' != originalEvent.data) {
+						//Make sure this is our event
+						return;
+					}
+					
+					if (searchOrigin != originalEvent.origin) {
+						//Verify the origin
+						return;
+					}
+					
+					closeSearch();
+				});
+
+				//Close search on escape
+				$(document).on('keydown', function(e) {
+					if (!e.keyCode || e.keyCode === 27) {
+						//Close on escape
+						closeSearch();
+					}
+				});
+
 				// listen for clicks on the document and hide the iframe if they didn't come from search interface
 				$(document).on('click', function(e) {
 					var $wdnSearch = domSearchForm.parent();
 					if (!$wdnSearch.find(e.target).length) {
-						$wdnSearch.removeClass('active');
+						closeSearch();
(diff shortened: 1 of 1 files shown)


wdn/templates_4.0/scripts/search.js
```diff
@@ -151,15 +156,49 @@ define(['jquery', 'wdn', 'require', 'modernizr', 'navigation'], function($, WDN,
...
+						//Verify the origin
+						return;
+					}
+					
+					closeSearch();
+				});
+
+				//Close search on escape
+				$(document).on('keydown', function(e) {
+					if (!e.keyCode || e.keyCode === 27) {
```
kabel: `!e.keyCode` ? What is that catching? Seems like this condition might catch more than it should.

wdn/templates_4.0/scripts/search.js
```diff
@@ -151,15 +156,49 @@ define(['jquery', 'wdn', 'require', 'modernizr', 'navigation'], function($, WDN,
 					if (postReady && $unlSearch[0].contentWindow.postMessage) {
 						e.preventDefault();
 						$unlSearch[0].contentWindow.postMessage(domQ.val(), searchOrigin);
+						if ('auto' != source) {
```
kabel: Just FYI, using type-strict comparisons ( `===` , `!==` ) are generally encouraged for performance and to avoid logical accidents.

wdn/templates_4.0/scripts/search.js
mfairchild365: Good question.  I saw it referenced in a few examples and thought I'd try to find a case where it was helpful.  I couldn't find a good case for it and forgot to take it out.

wdn/templates_4.0/scripts/search.js
mfairchild365: Good point

mfairchild365:
@kabel one question that I had: what is the best way to handle adding event listeners to DOM elements in a framework environment?

For example, in this PR I have:
```
$(window).on('message', function(e) {
```

Now, I could see legitimate cases where developers who use the framework might want to listen for their own iframe postMessage events in the same way.  What is the best way to ensure that our event handler is not ignored, and that it does not affect other event handlers?

kabel:
Strictly speaking there is nothing we can do to _prevent_ other developers from adding additional event listeners for the `postMessage` event (or removing, blocking this listener, for that matter). However, I believe the new handlers that are added are executed in the order that they were added. So as long as long a an earlier handler did not stop the event propagation, multiple handlers should lovingly coexist.

PR #916, opened 2015-12-04
Update Site Title, Navigation styles

skoolbus39:


wdn/templates_4.1/less/layouts/navigation.less
```diff
@@ -1,3 +1,5 @@
+@import (reference) "wrapper.less";
```
kabel: What is this? We should not be putting shared references in non-mixin files.

wdn/templates_4.1/less/layouts/navigation.less
```diff
@@ -112,10 +125,24 @@
         transition: height @nav-transition-duration ease;
         display: block;
 
+        @media @bp480 {
+            padding-left: @x-480;
```
kabel: I know I missed this previously, but I'm not a fan of these variable names. Using a dash makes it hard to distinguish the name vs math. Needs a meaningful name.

PR #918, opened 2015-12-07
auto close dropdown widgets

mfairchild365:
For now, this includes support for the mobile search and the share widget, but should be expanded to the IDM widget when implemented. It is built for to be plug and play with existing checkbox+label+css dropdown implementations.

In addition to only allowing one dropdown to be open a time, it improves a11y and closes dropdowns on `esc` or click outside of the dropdown.


wdn/templates_4.1/includes/share.html
```diff
@@ -1,13 +1,3 @@
...
-	    <label for="wdn_share_toggle" class="wdn-icon-share"><span class="wdn-text-hidden">Share This Page</span></label>
-	    <ul class="wdn-share-options">
-	        <li><a href="http://go.unl.edu/?url=referer" id="wdn_createGoURL" class="wdn-icon-link" rel="nofollow">Get a Go URL</a></li>
-	        <li class="outpost" id="wdn_emailthis"><a href="mailto:" class="wdn-icon-mail" rel="nofollow">Email this page</a></li>
-	        <li class="outpost" id="wdn_facebook"><a href="https://www.facebook.com/sharer/sharer.php" class="wdn-icon-facebook" rel="nofollow">Share on Facebook</a></li>
-	        <li class="outpost" id="wdn_twitter"><a href="https://twitter.com/share" class="wdn-icon-twitter" rel="nofollow">Share on Twitter</a></li>
-	        <li class="outpost" id="wdn_linkedin"><a href="https://www.linkedin.com/shareArticle?mini=true" class="wdn-icon-linkedin-squared" rel="nofollow">Share on LinkedIn</a></li>
-	    </ul>
-	</div>
+<div class="wdn-share-this-page wdn-dropdown-widget-content" id="wdn-main-share-button">
```
kabel: This essentially drops support for the share button in no-js browsers. Just any FYI.

mfairchild365: Are we actively supporting no-js use of the framework?

kabel: That was (what I assume) was the driving factor for using the `wdn-input-driver` elements. Otherwise, it's just element overkill (a label doing a button's job). The continued use of that pattern seems like support.

PR #926, opened 2015-12-15
Breadcrumb, navigation, color updates

skoolbus39:
Update UNL scarlet to #dd0000
Adjust layout of last-link breadcrumb
Adjust border-color for mobile navigation links
Position share widget on right edge of content instead of moving with navigations that have fewer than six primary links
Add border-bottom for primary navigation links without children


skoolbus39:
For clarification: the UNL scarlet was updated from #d00000 to #dd0000.

PR #934, opened 2016-01-04
Fix a11y tests

mfairchild365:
This should fix a11y tests that have been broken for 4.1.  (its a good thing I double checked them).

This also makes it easier to test without vagrant and improves the output of tests.

### DIFF
--- a/wdn/templates_4.1/less/layouts/footer.less
+++ b/wdn/templates_4.1/less/layouts/footer.less
@@ -165,6 +165,7 @@
 .wdn-footer-global {
 	padding: 3.157em 0 9.969em;
     color: @ui06;
+	background-color: @ui10;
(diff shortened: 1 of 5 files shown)


wdn/templates_4.1/less/layouts/footer.less
```diff
@@ -165,6 +165,7 @@
 .wdn-footer-global {
 	padding: 3.157em 0 9.969em;
     color: @ui06;
+	background-color: @ui10;
```
kabel: Why is this needed? It's already in the shorthand.

mfairchild365: See my commit message here: https://github.com/mfairchild365/wdntemplates/commit/eb9b4374586d14fc8a5646943193197052349e56

TLDR: phantomjs bug

PR #935, opened 2016-01-04
Remove vagrant

mfairchild365:
Vagrant was not very helpful for this project. It was most helpful for running the suite of tests, but those tests have been modified to allow testing on your local machine.


kabel:
Woot. The source code status are now more representative!

Issue #940, opened 2016-01-11
Make scroll depth tracking optional


mfairchild365:
Code already exists at https://github.com/unl/wdntemplates/blob/develop/wdn/templates_4.1/scripts/analytics_scroll_depth.js

Goals: Allow site owners to enable scroll depth tracking on a per page or per site basis. This should be optional and enabled with a code snippet.

Documentation for this feature should also be created.

mfairchild365:
Closed in favor of documentation. Found at http://wdn.unl.edu/ga-scroll-depth-tracking

PR #949, opened 2016-01-25
Add a jquery scrolldepth tracker

mfairchild365:
This will require some documentation, but I wanted to get this PR in for review first. Is there a better way to integrate this? We already have a scroll tracking solution, but it appears to be custom and I'm not even sure if it is being used.

This offers a good deal of customization, and can be turned only for sites with their custom GA tracking.

Library: http://scrolldepth.parsnip.io/

Example Code:
```
//ga tracking code here

//custom scroll depth tracking here
require(['jquery', 'plugins/scrolldepth/jquery.scrolldepth'], function($) {
  $.scrollDepth({
    //options
  });
});
```

#940

### DIFF
--- a/wdn/templates_4.1/scripts/plugins/scrolldepth/jquery.scrolldepth.js
+++ b/wdn/templates_4.1/scripts/plugins/scrolldepth/jquery.scrolldepth.js
@@ -1,0 +1,319 @@
+(function(factory) {
+	if (typeof define === 'function' && define.amd) {
+		// AMD
+		define(['jquery'], factory);
+	} else if (typeof module === 'object' && module.exports) {
+		factory(require('jquery'));
+	} else {
+		// Browser globals
+		factory(jQuery);
+	}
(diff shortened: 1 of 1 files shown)


cmdprompt:
I cast my vote to remove the scroll depth originally. It uses a huge amount of tracking events which have a limit. It was determined that people do in fact scroll on web pages. What additional information do people hope to get from this?

mfairchild365:
@cmdprompt you bring up a good point. This won't affect you in any way, as it has to be implemented on a per page/site basis. However, how many people would actually use this? Is it worth adding to the framework just for a few sites? Should we encourage using it?

I think there are use cases, some sites and content might not follow the trend, so site specific research is valuable.

The package doesn't work with requireJS out of the box, so I had to wrap it with some code to make it work. Perhaps the alternative to including it in the framework is to document on the WDN site how to add this package to your site (upload the library, modify it, use it).

Thoughts?

PR #955, opened 2016-02-09
Search: Fix an issue where http request was being switched to https

mfairchild365:
The first submit was fine, but after that, the origin would switch to https and thus break new submits. So if you started to type 'stephenie', it would submit a search for 'step' and then stop.


kabel:
Already fix at https://github.com/unl/wdntemplates/commit/6658ed56329673b67472db8d6663fb632bd248d2

Issue #957, opened 2016-02-11
Add room numbers (optionally) to events and events-band widgets


tlemburg:

tlemburg:
Pull request merged, closing

PR #958, opened 2016-02-11
Room number in Events and Events Band widget

tlemburg:
https://github.com/unl/wdntemplates/issues/957

Per request from College of Law. Many sites have more or less a single building hosting all their events, and would like to show the room number in their widgets.


tlemburg:
If possible, I'd like to get this into a point release to go out once it's tested and approved. But if we choose to wait until the next regular release, that's ok too.

kabel:
If we make a "fire" release for the existing framework issue (that has a known workaround), I'll be sure to get this in. Otherwise, it will have to wait for the next scheduled release on March 7.

PR #961, opened 2016-02-15
Update tests to work with root-less travis-ci

kabel:
Web server will be provided by a static http server from node and the
mod_include features will be handled in the PHP test.


tests/Accessibility/accessibility.phpt
```diff
@@ -12,40 +13,40 @@ if (file_exists(__DIR__ . '/../config.inc.php')) {
 class AccessibilityTester {
     protected $examples_directory = '';
     protected $wrapper_html;
-    public static $base_url = 'http://localhost/';
-    
+    public static $base_url = 'http://localhost:8080/';
```
mfairchild365: Is it possible to have the http-server use port 80?

kabel: No. Root is required to bind to ports < 1024.

PR #962, opened 2016-02-22
Upgrade MediaElement plugin to v2.19.0 from upstream

kabel:
Maintains patch from
https://github.com/johndyer/mediaelement/commit/0d5201490166ff77c79cdfc9685c0005caf05fbe


mfairchild365:
@kabel It looks like they have made a few releases since 2.19.0. We can tackle those for the next release.

PR #964, opened 2016-02-25
Improve iframe (desktop) search auto-submit timing

kabel:
Slow typing is causing too many queries on the search apps. This new
setup will immediately load/display the search iframe (with a loading
indicator) and only issue a submit after a full second of inactivity on
the keyboard.

### DIFF
--- a/wdn/templates_4.1/scripts/search.js
+++ b/wdn/templates_4.1/scripts/search.js
@@ -135,6 +135,7 @@ define(['jquery', 'wdn', 'require', 'modernizr', 'navigation'], function($, WDN,
 				};
 
 				var closeSearch = function() {
+					clearTimeout(autoSubmitTimeout);
 					domSearchForm.parent().removeClass('active');
 					$progress.hide();
 					domSearchForm[0].reset();
--- a/wdn/templates_4.1/scripts/search.js
+++ b/wdn/templates_4.1/scripts/search.js
@@ -1,4 +1,6 @@
 define(['jquery', 'wdn', 'require', 'modernizr', 'navigation'], function($, WDN, require, Modernizr, nav) {
+	var autoSearchDebounceDelay = 1000;
+
 	function getLocalSearch() {
 		var link = $('link[rel=search]');
 		if (link.length && link[link.length - 1].type != 'application/opensearchdescription+xml') {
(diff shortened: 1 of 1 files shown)


wdn/templates_4.1/scripts/search.js
```diff
@@ -99,29 +106,78 @@ define(['jquery', 'wdn', 'require', 'modernizr', 'navigation'], function($, WDN,
...
+
+						$unlSearch.on('load', function() {
+							postReady = true; // iframe should be ready to post messages to
+						});
+					}
+				};
+
+				var activateSearch = function() {
+					domSearchForm.parent().addClass('active');
+					$progress.show();
```
mfairchild365: `$progress.hide()` should now be added to the `close()` function. If you close the search before it finishes, a loading bar will be stuck on the screen.

wdn/templates_4.1/scripts/search.js
```diff
@@ -99,29 +106,78 @@ define(['jquery', 'wdn', 'require', 'modernizr', 'navigation'], function($, WDN,
...
 					clearTimeout(autoSubmitTimeout);
+
 					if ($(this).val()) {
+						// activate search UI
+						createSearchFrame();
+						activateSearch();
+
+						// debounce auto-submit
 						autoSubmitTimeout = setTimeout(function() {
 							domSearchForm.trigger('submit', 'auto');
```
mfairchild365: Should probably check if the search is still open. If it was closed between a keypress and auto-submit, it will pop back open.

wdn/templates_4.1/scripts/search.js
```diff
@@ -99,29 +106,78 @@ define(['jquery', 'wdn', 'require', 'modernizr', 'navigation'], function($, WDN,
...
-						}, 300);
+						}, autoSearchDebounceDelay);
```
mfairchild365: Should probably check if the search is still open. If it was closed between a keypress and auto-submit, it will pop back open.

wdn/templates_4.1/scripts/search.js
kabel: Added.

wdn/templates_4.1/scripts/search.js
kabel: I believe that is the expectation.The search _should_ come back up when submitted. See line 192.

mfairchild365: I disagree. If a user explicitly closes the search interface (either by clicking away from it or pressing escape) and it closes, it should not open itself.

mfairchild365: You are correct, that is currently the expectation in the code. However, the longer timeout makes the issue noticeable.

kabel: Or were you thinking that a close event would kill the next auto-submit? That is possible by clearing the timeout in the close function.

kabel: Clear added line 138.

PR #965, opened 2016-02-26
Use the directory avatar service

kabel:
This will bring in profile icons from the avatar services supported
through directory.


mfairchild365:
I think this is fine. It might be worth noting that this will increase load on directory.unl.edu. It might be worth caching the effective URL in directory's avatar service (but that has complications too).

PR #968, opened 2016-03-09
Move Mercury Display to optional plugin, add Sanelma script font plugin, update quote & infographics styles

### LINKED ISSUE #946: Move Mercury Display to optional, extended font set
With 4.1, we're no longer using Mercury Display for the site title. The only other framework asset (that I can think of) that uses Mercury Display is [quotes](http://wdn.unl.edu/wdn/templates_4.1/examples/#quote.html). By restyling the quotes and moving Mercury Display from the core font package to an optional, extended package, we could be saving at least 76KB.

This will require communication with the WDN Shared Governance Board and WDN.

### LINKED ISSUE #947: New quote style
Update [quote styling](http://wdn.unl.edu/wdn/templates_4.1/examples/#quote.html)

See issue #946

### LINKED ISSUE #967: Add Sanelma script font as extended font set

skoolbus39:
* Move Mercury Display to optional plugin
* Change Mercury Display in quote and infographics to Mercury ScreenSmart (Mercury Text) — See issue #946
* Reduce font-size and letter-spacing to account for difference in typeface proportions
* Adjust quotation mark spacing for better alignment
* Update code formatting to use parametric mixins
* Add Sanelma script font as optional Javascript plugin

Closes #946 Closes #947 Closes #967


wdn/templates_4.1/scripts/script-font.js
```diff
@@ -0,0 +1,25 @@
...
+            // protect against multiple initializations
+            if (initd) {
+                return;
+            }
+            initd = true;
+
+            // Load Sanelma script font
+            var $scriptFont = $("<link>", {
+                "rel" : "stylesheet",
+                "type" : "text/css",
```
kabel: As a coding standard, I recommend _not_ including the type attribute for stylesheets.

wdn/templates_4.1/scripts/script-font.js
```diff
@@ -0,0 +1,25 @@
...
+                "href" : "js-css/script-font.css"
```
kabel: I'm confused, this CSS file is already in the require statement. Why are you adding it again with this link element.?

kabel: I admit the difference to the other font plugins is subtle, but this is a different beast.
https://github.com/unl/wdntemplates/blob/72bcf10bf2cffa31a43c182d337d905d4b572725/wdn/templates_4.1/scripts/smallcaps.js#L19 represents an _external_ (vendor) CSS file that needs to be loaded into the head. https://github.com/unl/wdntemplates/blob/72bcf10bf2cffa31a43c182d337d905d4b572725/wdn/templates_4.1/scripts/smallcaps.js#L3 represents a _local_ (built) CSS file that will automatically be injected into the page when the script module is loaded.

They are indeed different files that serve a similar purpose. In this case, the local file contains an `@import` of the external file. So the need for a head appending link may not be necessary. I need to check to see how requirejs builds this CSS dependency to ensure it handles the import and relative font urls.

wdn/templates_4.1/scripts/js-css/script-font.less
```diff
@@ -0,0 +1,38 @@
...
+ * Licensed pageviews: 10,000,000
+ *
+ *
+ * License: http://www.myfonts.com/viewlicense?type=web&buildid=3193178
+ *
+ * © 2016 MyFonts Inc
+*/
+
+/* @import must be at top of file, otherwise CSS will not work */
+@import url("//hello.myfonts.net/count/30b95a");
```
kabel: Another coding standards practice I want to see adopted is _always_ using `https` when available, rather than the `//` scheme relative URLs. This has already been implemented in most other framework files.

kabel: As I feared. The LESS compiler will remove this import as it interprets it as a remote _less_ file. It should be replaced with `@import (css) url(https://hello.myfonts.net/count/30b95a);` to ensure the compiler copies it unmodified to the resulting css file.

wdn/templates_4.1/scripts/js-css/script-font.less
```diff
@@ -0,0 +1,38 @@
...
+
+@font-face {
+    font-family: 'Sanelma';
+    src: url('../../fonts/script-font/30B95A_0_0.eot');
```
kabel: Another fear about the RequireJS compiler. These URLs are outside of the "scripts" base and RequireJS will not allow that. We have a couple options that we can discuss in-person.

kabel: One more thing, if you go with the data uri approach we talked about, make sure you adjust the Gruntfile.js to pass the `ieCompat` flag to the `less:js` task. I noticed I didn't add that flag there yet.

wdn/templates_4.1/scripts/smallcaps.js
```diff
@@ -13,13 +13,12 @@ define([
             initd = true;
     
             // Load Mercury ScreenSmart Small Caps fonts from Cloud.typography
-            var $smcaps = $("<link>", {
-                "rel" : "stylesheet",
-                "type" : "text/css",
-                "href" : "//cloud.typography.com/7717652/679648/css/fonts.css"
+            var $smallCaps = $('<link>', {
+                'rel' : 'stylesheet',
+                'href' : '//cloud.typography.com/7717652/679648/css/fonts.css'
```
kabel: While you're working on the examples for these, please add the https.

PR #975, opened 2016-03-28
Add Modernizr tests for Video Autoplay, Object Fit

skoolbus39:
* The Video Autoplay test will help developers provide fallback content (usually in the form of poster images) for browsers (usually mobile) that don't support automatic playback of video content.
* The Object Fit test will allow developers to take advantage of the [object-fit property](https://dev.opera.com/articles/css3-object-fit-object-position/) for greater control over how images scale and fit inside parent element containers, and provide fallback layouts for the Microsoft browsers that don't support this useful property.


kabel:
Please keep in mind that `autoplay` is an "asynchronous" Modernizr test, so any javascript that relies on it should know how to listen for the result: https://github.com/Modernizr/Modernizr/blob/master/readme.md#new-asynchronous-event-listeners

PR #977, opened 2016-03-30
Remove some more network requests

kabel:
Removes 2 network requests for css related items (print sheet and large SVG image).


skoolbus39:
Looks like you're missing a new line at the end of your print.less file

kabel:
Nope. It's there.

skoolbus39:
Weird. When viewing a file in GitHub it doesn't show the final line. I never noticed that before.

PR #988, opened 2016-04-19
Bug fixes

mfairchild365:


wdn/templates_4.1/less/layouts/share.less
```diff
@@ -183,6 +183,18 @@
     }
 }
 
+/* the share widget dropdown needs some focus love because it's icon is absolutely positioned */
+.wdn-dropdown-widget-label.focused .wdn-icon-share {
+	/* firefox */
+	outline: 1px dotted #212121;
+	/* -webkit-focus-ring-color = '#5B9DD9' */
```
kabel: Can you explain this?

wdn/templates_4.1/scripts/navigation.js
```diff
@@ -665,6 +665,10 @@ define(['jquery', 'wdn', 'modernizr', 'require'], function($, WDN, Modernizr, re
 						switchSiteNavigation($homepageCrumbLink[0], false);
 					});
 
+					$(navSel + ' a').last().focusout(function(){
```
kabel: This will probably work for a majority of the cases, but it will fail with the navigation switcher (and probably the other event listeners that are bound the individual elements). It could be made better by binding the event to a "safe" parent element and only responding when the trigger matches the expectation. Make sense?

wdn/templates_4.1/less/layouts/share.less
mfairchild365: what about it? The icon is absolutely positioned, not the label (which usually gets the outline). So, when the control was focused, the label was being outlined but had no width (resulting in a line in some browsers, and no outline in others). This simply tries to simulate the default outline styles in firefox and webkit based browsers. 

This all stems from using the 'pure css' dropdown pattern, which utilizes a checkbox to track the visibility state. Browsers don't give focus to labels, so in order to provide a visual signal that you are current focused on something, we have to do this weird hack.

I'm all for ditching the 'pure css' dropdown pattern we have going on.

wdn/templates_4.1/scripts/navigation.js
mfairchild365: Gotcha. It will fail when the navigation switcher puts new navigation in the navigation container, thus dropping the old nav and the listener on the `<a>`. 

So you are saying add a listener to the parent `nav` container and listen to all `focusout` events, when the trigger is the last `<a>`, close the nav.

That makes sense, but it seems like a lot of overhead. I can't think of a better way of doing it.

wdn/templates_4.1/less/layouts/share.less
kabel: I meant explain what this line comment is for and why webkit hacks are necessary. Focus-like styles make perfect sense. 

But it you want to try for getting rid of the label/checkbox approach. I'm all for it. That approach was to support javascript disabled browsers. And that simply doesn't make sense for the social share that almost certainly requires JS.

mfairchild365: Ah, webkit has a different default outline color/style. This was an attempt to simulate it if it existed. I will be honest, this was a copy/paste from stack overflow. The comment is simply stating what the value of `-webkit-focus-ring-color` should be. If you want, the comment can be removed as it isn't super helpful.

wdn/templates_4.1/scripts/navigation.js
kabel: It's a _trivial_ amount of overhead; events are always bubbling and the more events we can condense into a single element reduces the memory footprint.

PR #993, opened 2016-05-12
Improve site title vertical alignment

skoolbus39:
Adjust site affiliation min-height (for sites with no site affiliation content present) to better vertically align the site title with the Nebraska 'N'.


kabel:
Do you mind adjusting the width of the N again so it's square ?

PR #994, opened 2016-05-13
Move font files from framework to top-level

skoolbus39:


kabel:
I do not suggest this for these files. As they are already designed to use data-uris.

skoolbus39:
After some discussion, we've decided to leave the script and icon fonts in the framework. I'm closing this pull request and will open a new one once we're ready to move the Cloud.typography fonts.

PR #999, opened 2016-05-20
Font styling updates

skoolbus39:
- Remove ligatures from core webfonts (used only by Mercury Display)
- Update kerning and ligature CSS for Mercury Display
- Remove 'font-kerning' declaration
- Remove redundant kerning declarations
- Remove sans serif font weight bold

### DIFF
--- a/wdn/templates_4.1/less/layouts/noscript.less
+++ b/wdn/templates_4.1/less/layouts/noscript.less
@@ -24,7 +24,6 @@
     background-color: @warning;
     color: #000000;
     .sans-serif-font();
-    font-weight: bold;
     text-align: center;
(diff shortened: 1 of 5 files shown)


wdn/templates_4.1/less/layouts/noscript.less
```diff
@@ -24,7 +24,6 @@
     background-color: @warning;
     color: #000000;
     .sans-serif-font();
-    font-weight: bold;
```
kabel: No that it matters for this pull request, but it should be documented that this base font _does not_ include a bold weight. The chances for people to accidentally use it are going to be common.

PR #1002, opened 2016-06-01
WIP: Improve a11y for tabs

mfairchild365:
This PR aims to improve a11y for tabs, specifically in regards to screen readers. It does the following:

- Applies appropriate roles to tab content such as `tablist`, `tab`, and `tabpanel`
- Replaces default keyboard navigation with the usual 'tab' navigation found in other UIs. In other words, it now uses arrow keys to cycle through tabs.

Issues that this solves
- SRs did not know that a list of links were 'tabs' or the relationship between tabs and tab panels
- SRs did not know which tab was currently selected
- SRs did not know the boundaries of tab panels (tab content)
- SR users that are familiar with tab UIs in non-browser environments expect to navigate the tabs with arrow keys. (this is news to me and there has been some debate around this, but I will defer to the experts).

Specifically this follows best practices as described by https://medium.com/@LeonieWatson/danger-testing-accessibility-with-real-people-4515f72db648 and http://a11y.nicolas-hoffmann.net/tabs/

This is a work in progress, but it is in a state that is appropriate for review. I still need to do some more testing before it gets merged. Feedback is welcome.

### DIFF
--- a/wdn/templates_4.1/scripts/tabs.js
+++ b/wdn/templates_4.1/scripts/tabs.js
@@ -101,7 +101,7 @@ define(['jquery', 'wdn', 'require'], function($, WDN, require) {
 				var $panel = $('#'+getHashFromLink(this));
 
 				//If this tab doesn't have an id, give it one because we will need it later
-				if ('undefined' === typeof $tab.attr('id')) {
+				if (!this.id) {
 					$tab.attr('id', 'wdn-tab-for-'+$panel.attr('id'));
 				}
 
--- a/wdn/templates_4.1/scripts/tabs.js
+++ b/wdn/templates_4.1/scripts/tabs.js
@@ -94,12 +94,11 @@ define(['jquery', 'wdn', 'require'], function($, WDN, require) {
 
 			var $tabs = $($tabsWithSwitch.find('a'));
 
-			$tabs.each(function(index, element) {
-				console.log(element);
-				var $tab = $(element);
+			$tabs.each(function() {
+				var $tab = $(this);
 
 				//Get the panel for this tab
-				var $panel = $('#'+getHashFromLink(element));
+				var $panel = $('#'+getHashFromLink(this));
 
 				//If this tab doesn't have an id, give it one because we will need it later
 				if ('undefined' === typeof $tab.attr('id')) {
--- a/wdn/templates_4.1/scripts/tabs.js
+++ b/wdn/templates_4.1/scripts/tabs.js
@@ -45,6 +45,7 @@ define(['jquery', 'wdn', 'require'], function($, WDN, require) {
 				nsel;
 
 			// Remove any selected tab class
+			$('li.' + selected, tabs).find('a').attr('tabindex', '-1').attr('aria-selected', 'false');
 			$('li.' + selected, tabs).removeClass(selected);
 
 			// Hide any subtabs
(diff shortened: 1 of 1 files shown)


wdn/templates_4.1/scripts/tabs.js
```diff
@@ -64,6 +66,70 @@ define(['jquery', 'wdn', 'require'], function($, WDN, require) {
...
+				$tab.attr('aria-selected', 'false');
+
+				//Tell the tab that it controls this panel
+				$tab.attr('aria-controls', $panel.attr('id'));
+				
+				//Only make tabs pragmatically focusable by default
+				$tab.attr('tabindex', '-1');
+				
+				//Use arrow keys to focus tabs (standard tab interaction)
+				$tab.bind({
```
kabel: Please don't use the bind method. I recommend adding this event listener to the `$tabsWithSwitch` and using delegation if necessary. See `.on()` API documentation.

wdn/templates_4.1/scripts/tabs.js
```diff
@@ -64,6 +66,70 @@ define(['jquery', 'wdn', 'require'], function($, WDN, require) {
...
+					keydown: function(ev){
+						var k = ev.which || ev.keyCode;
```
kabel: No need to do the `||` here, as jQuery already normalizes this browser difference. Also the variable name should be better.

wdn/templates_4.1/scripts/tabs.js
```diff
@@ -64,6 +66,70 @@ define(['jquery', 'wdn', 'require'], function($, WDN, require) {
...
 			trig.trigger('tabchanged', [curr, nsel, tabs]);
 		},
+		
+		setupA11y = function ($tabsWithSwitch) {
+			$tabsWithSwitch.attr('role', 'tablist');
+
+			var $tabs = $($tabsWithSwitch.find('a'));
+
+			$tabs.each(function(index, element) {
+				console.log(element);
```
kabel: No debug statements in the merge, please.

wdn/templates_4.1/scripts/tabs.js
```diff
@@ -64,6 +66,70 @@ define(['jquery', 'wdn', 'require'], function($, WDN, require) {
 			nsel = trig.closest('li').siblings('.' + selected);
 			trig.trigger('tabchanged', [curr, nsel, tabs]);
 		},
+		
+		setupA11y = function ($tabsWithSwitch) {
+			$tabsWithSwitch.attr('role', 'tablist');
+
+			var $tabs = $($tabsWithSwitch.find('a'));
+
+			$tabs.each(function(index, element) {
```
kabel: Unless index is used, it's easier just to use no arguments in the callback and rely on the callback context `this` being the element.

wdn/templates_4.1/scripts/tabs.js
mfairchild365: Hey man, its a WIP. :P But I appreciate the note.

wdn/templates_4.1/scripts/tabs.js
mfairchild365: TIL

wdn/templates_4.1/scripts/tabs.js
```diff
@@ -64,6 +66,70 @@ define(['jquery', 'wdn', 'require'], function($, WDN, require) {
...
+				console.log(element);
+				var $tab = $(element);
+
+				//Get the panel for this tab
+				var $panel = $('#'+getHashFromLink(element));
+
+				//If this tab doesn't have an id, give it one because we will need it later
+				if ('undefined' === typeof $tab.attr('id')) {
```
kabel: `if (!this.id)` might be better in this case as any empty id would be bad.

mfairchild365:
Note: there is a safari/voiceover bug which renders this implementation useless. See http://mfairchild365.github.io/aria-tabs-vo-bug/

kabel:
Closing until such a time that these will actually help an agent.

PR #1010, opened 2016-07-18
WIP: A11y improvements

mfairchild365:
NOTE: this PR is under development, do not merge yet. I've created this PR as a place to discuss and coordinate.

This aims to fix the following issues:

- [x] - [dropdown widget controls (search, share, idm)] Convert the input-label style of buttons to just `<button>` elements. The css only input-label style has several a11y issues associated with it, including incorrect semantics, bad focus-management, inconsistent outlines, etc. I see little benefit to using the css-only trick here because we will need to use JS anyway (for aria) and the framework requires JS to work anyway.
 - [x] - [MF] update wdn-ui.js to reflect changes
 - [x] - [MF] update includes to reflect changes
 - [x] - [RD] update styles to hide/show containers based on the container's `aria-hidden` attribute.
- [x] - [MF] [Navigation controls] replace the input-label style of buttons for the navigation toggle as well. This is slightly different than the dropdown widget controls and will require some extra consideration.

In a future release: Improve tab order of these controls as well. Because they are moved around with css at different widths, tab order is sometimes incorrect.


wdn/templates_4.1/scripts/wdn-ui.js
```diff
@@ -56,7 +56,7 @@ define(['jquery', 'modernizr'], function($, Modernizr) {
 				
 				//Try to get the control element
 				if ($control.parent('.'+dropdownButtonClass).length) {
-					$control = $control.parent('.'+dropdownButtonClass)[0];
+					$control = $($control.parent('.'+dropdownButtonClass)[0]);
```
kabel: Not sure why you didn't just use the jQuery API `.eq` or `.first`

wdn/templates_4.1/scripts/main-wdn-plugins.js
```diff
@@ -1,8 +1,8 @@
 define([
 	'analytics',
+	'socialmediashare',
 	'navigation',
 	'search',
-	'socialmediashare',
```
kabel: These are all async and order does not matter. If `navigation` depends on `socialmediashare` the latter should be in the former's dependency definition.

mfairchild365: I thought order does matter because of the for loop with `pluginObj.initialize();`?

kabel: Sure, but you _should not_ depend on that. Dependencies should be explicitly declared and initialization should be entirely independent of the status of another module.

mfairchild365:
@kabel I think this is ready to be reviewed (I know its a lot).

I'm a little worried about the potential case where some sites might not have the newest version of the includes, but DO have the newest assets. I've added some CSS to hide the old labels, but some of the functionality won't work until the includes are updated. Would it be worth it to go a step further and make sure that the functionality works even if the includes are not updated? I'm guessing not.

Do you think this is okay to roll out?

wdn/templates_4.1/less/layouts/navigation.less
```diff
@@ -252,6 +254,10 @@
             }
         }
     }
+    
+    label {
+        display: none; /*BC fix, prevent duplicate nav buttons after change to buttons if includes are not yet updated */
```
kabel: Please use less comments to avoid adding comments to the compressed output.

wdn/templates_4.1/scripts/navigation.js
```diff
@@ -588,10 +591,62 @@ define(['jquery', 'wdn', 'modernizr', 'require'], function($, WDN, Modernizr, re
...
+		//Remove the nav input
+		//$(menuTogSel).hide();
+
+		//handle the label (this should be a button when sends focus)
+		var $navToggleButton = $('<button>');
+		$navToggleButton.html('<span class="wdn-icon-menu" aria-hidden="true"></span><span class="wdn-text-hidden">Menu</span>'); //Make sure they have the same HTML contents
+		$navToggleButton.addClass('wdn-nav-toggle');
+
+		//Handle click events
+		$([$navToggleButton]).each(function(index, $button) {
```
kabel: This seems weird. Why are you calling each? You already have a jQuery object, so you can go straight to `on`.

wdn/templates_4.1/scripts/wdn-ui.js
```diff
@@ -35,28 +39,47 @@ define(['jquery', 'modernizr'], function($, Modernizr) {
...
-				var $target = $(e.target);
+				//Determine if we need to do anything with our dropdown
+				var $control = $(e.target);
 
-				//dont close this way if the control was clicked
-				if ($target.filter('.wdn-dropdown-widget-toggle, .wdn-dropdown-widget-label').length) {
-					return;
+				//Try to get the control element
+				if ($control.parent('.'+dropdownButtonClass).length) {
+					$control = $control.parent('.'+dropdownButtonClass).first();
```
kabel: Are you sure you're looking for `parent` (only first parent)? Perhaps you mean `parents` or `closest`. `parent` will return 1 or 0 elements, so no need for first.

wdn/templates_4.1/scripts/wdn-ui.js
```diff
@@ -35,28 +39,47 @@ define(['jquery', 'modernizr'], function($, Modernizr) {
...
+					}
+
+					//Close other widgets
+					closeDropDown($('.'+dropdownButtonClass).not($control));
 				}
 
 				//close all dropdown widgets
-				if (!$('.wdn-dropdown-widget-content').find(e.target).length) {
-					closeDropDown('.wdn-dropdown-widget-toggle');
+				if (!$('.'+dropdownContentClass).find(e.target).length) {
```
kabel: Where we can, we should reduce the number of times to go to the selector engine with just a class.

wdn/templates_4.1/scripts/navigation.js
mfairchild365: I think this was a leftover from when we were doing something a little differently. I will improve it.

wdn/templates_4.1/less/layouts/navigation.less
mfairchild365: good suggestion!

mfairchild365:
@kabel I've added better support for backwards compatibility (if includes haven't been updated yet).

PR #1012, opened 2016-08-09
Update styles to hide/show containers based on the container's aria-hidden attribute 

skoolbus39:
Work still needs to be done to clean up the wdn-ui.js file and and change the aria-pressed state to false when clicking/tapping outside of share option popup.

### DIFF
--- a/wdn/templates_4.1/less/states/idm.less
+++ b/wdn/templates_4.1/less/states/idm.less
@@ -47,11 +47,11 @@
 		@media @bp768 {
             padding-right: 50px;
 		}
-	}
 
-    #wdn_idm_toggle:checked ~ .wdn-idm-options {
-        .is-visible();
-        transform: scale(1);
-        transition: opacity 0.1s ease-in-out, transform 0.1s ease-in-out;
-    }
+        &[aria-pressed='true'] ~ .wdn-idm-options {
+            .is-visible();
+            transform: scale(1);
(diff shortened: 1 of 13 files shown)


wdn/templates_4.1/less/states/search.less
```diff
@@ -6,15 +6,15 @@
...
+            top: 0;
             bottom: auto;
         }
 
         .nav-scrolling #wdn_menu_toggle:checked ~ #header & {
             left: 100%;
         }
 
-        .wdn_search_toggle:checked ~ #wdn_search_form {
+        .wdn-dropdown-widget-button[aria-pressed='true'] ~ #wdn_search_form {
```
kabel: Less probably fixes this, but this is not a valid selector. Single quotes cannot be used in attribute selectors.

wdn/templates_4.1/less/states/idm.less
```diff
@@ -47,11 +47,11 @@
...
             padding-right: 50px;
 		}
-	}
 
-    #wdn_idm_toggle:checked ~ .wdn-idm-options {
-        .is-visible();
-        transform: scale(1);
-        transition: opacity 0.1s ease-in-out, transform 0.1s ease-in-out;
-    }
+        &[aria-pressed='true'] ~ .wdn-idm-options {
```
kabel: Same issue here with single quotes not being valid.

PR #1014, opened 2016-08-26
Update example to encourage https

tlemburg:
HTTPS is getting to be required for Safari and IE sometimes, so let's stick that in the example to show users what to do.


kabel:
Just so we also have some background as to why this is happening. The production server is sending the `Strict-Transport-Security` header (meaning that domain should always and only connect via https). Without a permanent redirect from the non-secure scheme to the secure scheme, people that visit the secure version will break the widgets and pages that try to manually refer to the non-secure scheme.

PR #1018, opened 2016-09-26
Print style cleanup and restore logo styles

### LINKED ISSUE #1015: 'N' missing on first print attempt
Right now, its a background image that doesn't print. This appears to be where it was removed:

https://github.com/unl/wdntemplates/commit/534d5fbb708e6b677afdbd6508eb8bc51ab6a20f#diff-4b6f632b0d8a9f7362178e6cedf8caf6L110

skoolbus39:
Fixes #1015


kabel:
Revert back to using a logo url this will cause problems on the _first_ print when the image is not cached. The print will not include the logo until the logo is loaded. I think the only solution, if you want to use the generated content hack, is to use a data-uri with the understanding that is make the sheet a little bigger.

kabel:
Cherry-picked this into develop.

PR #1019, opened 2016-10-03
Increase button line-height

skoolbus39:
Line-height value of 1 is too tight for buttons with two lines of text. Because the buttons are set in uppercase Gotham, the line-height can be tight, but not too tight.

### DIFF
--- a/wdn/templates_4.1/less/modules/forms.less
+++ b/wdn/templates_4.1/less/modules/forms.less
@@ -31,7 +31,7 @@
     text-decoration: none;
     text-transform: uppercase;
     .wdn-sans-serif();
-    line-height: 1;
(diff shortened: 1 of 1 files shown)


skoolbus39:
Hold off on merging this at the moment. This will need more testing.

kabel:
Can you let me know what needs more testing and for how long you need to hold. I don't like to have indefinite pull request open.

skoolbus39:
@kabel, the improved line-height works great for button text, but doesn't work so well for form buttons that are icons only (e.g. the magnifying glass in the site search). I want to adjust these types of buttons and make sure that this update doesn't have other, similar unintended consequences. Now that other projects are winding down, I'd like to return to this ASAP.

PR #1020, opened 2016-10-04
Improve layout of hidden regions

skoolbus39:
Corrected layout issues for [regions that may be conditionally hidden](http://wdn.unl.edu/unledu-41-style-book#exceptions)

### DIFF
--- a/wdn/templates_4.1/less/states/regions.less
+++ b/wdn/templates_4.1/less/states/regions.less
@@ -99,7 +99,7 @@
 
 .hide-wdn_copyright { // Keep the UI presentation, but hide the content.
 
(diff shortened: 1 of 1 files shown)


mfairchild365:
@skoolbus39, @kabel pointed out that the navigation button is still shown in the case where `#navigation > ul` does not exist (no site navigation). It might be good to adjust the styles to only show the button when the `.nav_ready` class is added to `#wdn_wrapper`. Thoughts?

kabel:
This is what used to protect the label from showing: https://github.com/unl/wdntemplates/blob/develop/wdn/templates_4.1/less/states/navigation.less#L77-L85

PR #1033, opened 2016-11-01
Mystery grid addition

cmdprompt:
There are only 3 columns in the global footer. There is never a reason to make the spacing strange by switching to a 4 col layout with only 3 cols.


kabel:
I believe this was a purposeful decision to leave space for the seal and an active chat without overlap on desktop. I'll defer to @skoolbus39 for confirmation.

cmdprompt:
I see.

skoolbus39:
As both Kevin and I have explained (Kevin, here; and I, in our email exchange), the design of the global footer was indeed intentional, leaving space for an open chat window. You're welcome to use the local footer example markup I sent to better align the two footers. I will also take this into consideration as we update the design of the footer(s) in the future. We're never finished.

PR #1034, opened 2016-11-01
Update mediaelement

mfairchild365:


mfairchild365:
There seems to be an issue with the captions toggle now. :(  It still works if you click the button, but on hover it opens the captions selector box.

PR #1035, opened 2016-11-02
Add accessibility checking via axe-core

mfairchild365:
WebAudit will be moving to aXe core. For at least a month during the transition, both metrics will be running, so we should also run both metrics here. Once WebAudit drops pa11y, we can drop pa11y here too.


mfairchild365:
yey, now tests will take twice as long! /s

PR #1040, opened 2016-12-13
Remove rem/px function, update px/em/rem values

### LINKED ISSUE #1030: Rid the less of the rem mixin
It contributes to the following problems:
- Different modular scales when two pixel value parameters are passed
- Contains a media query that contributes to nested media query junk
- Parameters and fallback are in pixel units (obfuscates modular scale)

These issues alone contribute ~5K (or 1.8%) of cruft to the CSS.

skoolbus39:
This removes the function that generated rem and pixel fallback values for font-size when one integer (the same value for both 'mobile' and 'desktop') or two integers (one for 'mobile,' one for 'desktop') were inputted into the function. This function's primary purpose was to provide pixel fallback values for browsers that didn't yet support rem units. All of the browsers supported by the framework now support rem units. This function is no longer needed.

Closes #1030


kabel:
An obvious change that occurred is the search box on desktop got considerably wider. Is that intended? Will it impact placement of IdM and affiliate institution text.

PR #1046, opened 2017-01-24
Add vertical space between main content and footer

skoolbus39:
Add margin-bottom to last-child element (if not a .wdn-band) of #maincontent to add vertical space between main content and footer.

If someone can think of a better way of doing this, I'm all ears.

### DIFF
--- a/wdn/templates_4.1/less/layouts/bands.less
+++ b/wdn/templates_4.1/less/layouts/bands.less
@@ -1,6 +1,14 @@
-// implement the same content padding
-#maincontent > * {
-	.inner-wrapper-margin();
+#maincontent {
+
+    // implement the same content padding
+    > * {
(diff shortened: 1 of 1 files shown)


wdn/templates_4.1/less/layouts/bands.less
```diff
@@ -1,6 +1,14 @@
...
-	.inner-wrapper-margin();
+#maincontent {
+
+    // implement the same content padding
+    > * {
+	    .inner-wrapper-margin();
+    }
+
+    // Add margin to last element (if not a .wdn-band) to add vertical space between main content and footer
+    > *:last-child:not(.wdn-band) {
```
kabel: This probably could have been done without using a separate nesting level by way of `&`, but it's okay.

PR #1050, opened 2017-02-06
Update pixel fitting of Nebraska N logos

skoolbus39:
* Remove old Nebraska N SVG file (located in header)
* Replace with two SVG files, each pixel fitted for respective breakpoints
* Update Nebraska N white logo (located in footer)

nebraska-n-before-after


skoolbus39:
[Pixel-fit Nebraska N to the size we’re using in framework (current has blurred vertical outlines on non-retina displays)](https://trello.com/c/f4XUZMkX/40-pixel-fit-nebraska-n-to-the-size-we-re-using-in-framework-current-has-blurred-vertical-outlines-on-non-retina-displays)

PR #1052, opened 2017-02-16
 carousel: Disable The direction nav by default

mfairchild365:
The direction nav fails accessibility audits by default because in its 'hidden' state, the navigation is still in the DOM and focusable. However, when focused with a keyboard, they are not shown, and thus can cause confusion (where is the current focus?). Additionally, the animation to show them on hover floats them in from the sides and can create a confusing and moving click target. Let's just disable them for now.

### DIFF
--- a/wdn/templates_4.1/scripts/carousel.js
+++ b/wdn/templates_4.1/scripts/carousel.js
@@ -18,9 +18,15 @@ define([
 	return {
 		initialize : function(callback) {
 			$(function() {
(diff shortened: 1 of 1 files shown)


mfairchild365:
[Make the framework carousel pass axe by default at desktop widths](https://trello.com/c/tcIq0G4U/61-make-the-framework-carousel-pass-axe-by-default-at-desktop-widths)

PR #1053, opened 2017-02-16
remove margin to last element

cmdprompt:
This is causing white bars to appear on our site where their shouldn't be.

### DIFF
--- a/wdn/templates_4.1/less/layouts/bands.less
+++ b/wdn/templates_4.1/less/layouts/bands.less
@@ -4,11 +4,6 @@
     > * {
 	    .inner-wrapper-margin();
     }
-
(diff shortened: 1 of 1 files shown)


cmdprompt:
image

kabel:
The margin is getting added because of your `#aspnetForm` not having `.wdn-band` or adding to your CSS overrides for `#maincontent > form, #maincontent > .main-form`. 

The goal of this CSS is to prevent content not in bands from accidentally bumping into the bottom of the page. Something that was very common for pages that haven't converted to use these new layout classes.

cmdprompt:
Is your suggested solution to put `.wdn-band` on that form then?

kabel:
That, or change the specificity of your styles selector (add the `#aspnetForm` prefix to the second part), or add `! important` to your `margin` property.

PR #1055, opened 2017-03-07
Force events widgets to use https to retrieve events

tlemburg:
Safari, mobile Safari, and occasionally IE/Edge (some versions)
would not correctly upgrade requests to the UNL Events API
to https. This would cause a CSRF warning and the request would
fail, causing the events widget not to appear. The problem was
fixed by the user entering https:// before their calendar URL
as opposed to just //, but this patch forces https:// if the
user had the old way put in.

Fixes card
https://trello.com/c/7c3VGIvp/60-events-widget-force-https-urls-for-parameters-in-the-widget

### DIFF
--- a/wdn/templates_4.1/scripts/events-band.js
+++ b/wdn/templates_4.1/scripts/events-band.js
@@ -19,7 +19,7 @@ define([
     },
 
     container = '#events-band',
-    defaultCal = '//events.unl.edu/';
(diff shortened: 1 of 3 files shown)


mfairchild365:
Looks good! However, monthwidget.js should probably be updated too.

tlemburg:
I force pushed (like a Sith) an edit to monthwidget on top of this.

PR #1058, opened 2017-03-21
WIP: Add a speedy template

mfairchild365:
This PR adds a 'speedy.shtml' template which aims to boost page speed when compared to the normal 'fixed.shtml' template. I will attempt to explain design decisions and rational in this PR. Collaboration and feedback is more than welcome.

## Performance issues with the 'fixed.shtml' template

As Google [PageSpeed insights indicates](https://developers.google.com/speed/pagespeed/insights/), the framework has the following problems:

1. Framework javascript is render blocking. By render blocking, I mean that all.js is in the head of the document and prevents further page execution until it has been loaded and processed.
2. Framework CSS is render blocking. The all.css file is also included in the head of the page which prevents further processing of the page until all.css has been loaded and processed.

These two problems together might lead a slow page load. 

## Why a new template?

The only way to address the render blocking JavaScript issue is to move it to the bottom of the page so that it loads after everything else has loaded. Unfortunately, this means that any custom javascript that relies on the framework's all.js (`require`, `WDN`, plugins, etc) will no longer work because those javascript objects do not exist yet. This is a breaking change, and would require developers to move their custom javscript to the bottom of the page, below the all.js load, in a new editable region named `bodyscripts`.

So, we have two options. 1) release a new version of the framework such as 4.2, or 2) implement a new template that developers can opt into. I think option 2 is the best immediate approach, and would allow us to make the new speedy template the default template in the 4.2 release of the framework.

## CSS loading

To address the render blocking CSS issue, @skoolbus39 has created a critical-css package, which contains a minimal amount of CSS that can be inlined in the head of the document and thus loaded immediately. Later after above-the-fold content has been loaded, all.css will come in and the critical-css styles will be removed. To further improve performance, future page-loads will know that `all.css` is in the cache, and will thus be loaded ASAP (this is done via sessionStorage). Assuming the css is cached in this method may not be fool proof, but the worse-case-scenario is that a new version of the framework would be released and the cache busting query string on all.css (`?dep=4.1.x`) would force the css to be reloaded.

## Considerations

In developing this, I have kept the following goals at heart:

1) This should work across implementations (no additional server side logic is required to boost performance). For example, one possible solution to the CSS issue would have been to have web developers remove the critical css and replace it with all.css on subsequent page loads.
2) Provide as much backwards compatibility as possible. The core HTML structure of page has not changed, and neither has any of the JS. The only thing that has changed is WHEN the JS and CSS is loaded.

## TODO:

If we move forward with this, we will have to do the following:

- [x] actually implement critical.less (just minimized CSS right now for testing)
- [x] mobile is good, but desktop could still use some work (not sure how to address this without breaking up all.js). I think this can be further improved by adding to the critical css.
- [ ] Testing across browsers
- [ ] Test it on an actual site (test it on a development version of the top level site to identify potential issues)
- [ ] Create a speedy dwt template file (the php wdn template package would also need to be updated to include the new template)
- [x] Possibly improve the JS that I added

## WE DID IT!
Before: 
screen shot 2017-03-21 at 11 01 53 am

After:
screen shot 2017-03-21 at 11 02 08 am

## Call to action

Please provide feedback, ideas, and help collaborate on this.


speedy.shtml
```diff
@@ -0,0 +1,139 @@
...
+<head>
+    <!--#include virtual="/wdn/templates_4.1/includes/metanfavico.html" -->
+    <!--
+        Membership and regular participation in the UNL Web Developer Network is required to use the UNLedu Web Framework. Visit the WDN site at http://wdn.unl.edu/. Register for our mailing list and add your site or server to UNLwebaudit.
+        All framework code is the property of the UNL Web Developer Network. The code seen in a source code view is not, and may not be used as, a template. You may not use this code, a reverse-engineered version of this code, or its associated visual presentation in whole or in part to create a derivative work.
+        This message may not be removed from any pages based on the UNLedu Web Framework.
+    
+        $Id: speedy.dwt | 6edb0e1ee94038935f3821c6ce15dfd5c217b2e2 | Tue Dec 1 17:08:56 2015 -0600 | Kevin Abel  $
+    -->
+    <!--#include virtual="/wdn/templates_4.1/includes/scriptsandstyles_speedy.shtml" -->
```
kabel: I advise against using an `shtml` include here. While server-side includes _can_ be nested like this, the files we put in the `includes` directory must be synced across a number of platforms and server technologies. SSI is not likely to be a technology that is widely supported and deployed.

wdn/templates_4.1/includes/scriptsandstyles_speedy.shtml
```diff
@@ -0,0 +1,42 @@
...
+</style>
+<script>
+    var WDNRemoveCriticalStyles = function() {
+        document.querySelector('#wdn-critical-styles').remove();
+    };
+    var WDNLoadDeferredStyles = function() {
+        //Define sheets to load
+        function addSheet(url, onload) {
+            var sheet = document.createElement("link");
+            sheet.type = "text/css";
```
kabel: No need for type here.

wdn/templates_4.1/includes/scriptsandstyles_speedy.shtml
```diff
@@ -0,0 +1,42 @@
+<noscript>
+    <link rel="stylesheet" href="https://cloud.typography.com/7717652/616662/css/fonts.css" />
+    <link rel="stylesheet" href="https://unlcms.unl.edu/wdn/templates_4.1/css/all.css?dep=$DEP_VERSION$" />
+</noscript>
+<style id="wdn-critical-styles">
+    <!--#include virtual="/wdn/templates_4.1/css/critical.css" -->
+</style>
+<script>
+    var WDNRemoveCriticalStyles = function() {
```
kabel: I'd wrap all of this in an self-executing anonymous function to avoid polluting the global object with single use functions:
```
(function() {
...
})();


```

wdn/templates_4.1/includes/scriptsandstyles_speedy.shtml
```diff
@@ -0,0 +1,42 @@
...
+        document.querySelector('#wdn-critical-styles').remove();
```
kabel: If not doing a complex selector, it's best to use `document.getElementById` to avoid the overhead of parsing the selector.

wdn/templates_4.1/includes/scriptsandstyles_speedy.shtml
```diff
@@ -0,0 +1,42 @@
...
+        //Load the fonts
+        addSheet('https://cloud.typography.com/7717652/616662/css/fonts.css');
+        //Load the all the css and then remove the inline critical css
+        addSheet('https://unlcms.unl.edu/wdn/templates_4.1/css/all.css?dep=$DEP_VERSION$', WDNRemoveCriticalStyles);
+       
+        //Set a flag that will persist across page loads (until the browser is closed)
+        sessionStorage.setItem('wdn-speedy-assets-are-cached', true);
+    };
+    //Load the CSS
+    if (sessionStorage.getItem('wdn-speedy-assets-are-cached')) {
```
kabel: I'd put this magic key value `wdn-speedy-assets-are-cached` into a variable/constant that is used in both getting and setting. Change once.

wdn/templates_4.1/includes/scriptsandstyles_speedy.shtml
```diff
@@ -0,0 +1,42 @@
+<noscript>
+    <link rel="stylesheet" href="https://cloud.typography.com/7717652/616662/css/fonts.css" />
+    <link rel="stylesheet" href="https://unlcms.unl.edu/wdn/templates_4.1/css/all.css?dep=$DEP_VERSION$" />
+</noscript>
+<style id="wdn-critical-styles">
+    <!--#include virtual="/wdn/templates_4.1/css/critical.css" -->
```
kabel: The includes should remain static, server tech independent files. For the reasons mentioned above about how our sync process works, I would not put a SSI statement in this file, otherwise no other server technology (php, jsp, etc) will be able to use it. 

Another idea for this might be to use add a set in the build process that builds this include file from a prefix file, the built css file, and a suffix file.

wdn/templates_4.1/includes/scriptsandstyles_speedy.shtml
```diff
@@ -0,0 +1,42 @@
...
+            sheet.rel = "stylesheet";
+            sheet.href = url;
+            sheet.onload = onload;
+            document.getElementsByTagName("head")[0].appendChild(sheet);
+        }
+        
+        //Load the fonts
+        addSheet('https://cloud.typography.com/7717652/616662/css/fonts.css');
+        //Load the all the css and then remove the inline critical css
+        addSheet('https://unlcms.unl.edu/wdn/templates_4.1/css/all.css?dep=$DEP_VERSION$', WDNRemoveCriticalStyles);
```
kabel: You'll want to make sure this file is also smudged so the version keyword get replaced. I also don't know what effect this will have on the analytics tracking of the dependents version.

wdn/templates_4.1/includes/scriptsandstyles_speedy.shtml
```diff
@@ -0,0 +1,42 @@
...
+        //Assets are cached, so we should load right now (before JS is loaded)
+        WDNLoadDeferredStyles();
+    } else {
+        //Otherwise, load deferred
+        var raf = requestAnimationFrame || mozRequestAnimationFrame ||
+            webkitRequestAnimationFrame || msRequestAnimationFrame;
+        if (raf) raf(function() { window.setTimeout(WDNLoadDeferredStyles, 0); });
+        else window.addEventListener('load', WDNLoadDeferredStyles);
+    }
+</script>
```
kabel: N/L!

wdn/templates_4.1/includes/scriptsandstyles_speedy.shtml
kabel: If `WDNRemoveCriticalStyles` is only used here, I'd just use it here as an anonymous function.

mfairchild365:
@kabel you are MUCH too kind to review this. I very much appreciate your feedback.

skoolbus39:
Seconded. Thanks a bunch, @kabel.

PR #1059, opened 2017-03-27
Fix: Adding focus for events-band

Shaderpixel:
Outline was set to zero for events-band links (cards) with no alternative styles provided for when they are in focus. [Resource](http://www.outlinenone.com/)

-Added focus styles to event band links including view more events link at the bottom of the band.
-Added two color variables in events-band.less: @events-darkblue, and @events-darkbluegray

Changes to be committed:
	modified:   wdn/templates_4.1/scripts/js-css/events-band.less

### DIFF
--- a/wdn/templates_4.1/scripts/js-css/events-band.less
+++ b/wdn/templates_4.1/scripts/js-css/events-band.less
@@ -89,7 +89,6 @@
         }
 
         .dateTime {
-            background-color: @events-darkblue;
             background-color: fadeout(#000, 60%);
             float: left;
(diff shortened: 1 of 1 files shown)


skoolbus39:
The fallback hex background-color for .dateTime is no longer needed. RGBA color is now supported by all modern browsers. @Shaderpixel, can you remove?

Shaderpixel:
@skoolbus39 It's ready for another review. Thanks.

Shaderpixel:
@skoolbus39 Requested changes addressed.

PR #1063, opened 2017-04-21
Speedy updates

skoolbus39:
I've made some improvements that put us in the green overall, but work remains in a few areas. 

Given that loading Cloud.typography web fonts requires a handshake with their servers on each page load, our optimization options are rather limited. The best approach I've found so far has been to use a combination of `rel="preload"` and `onload="this.rel='stylesheet'"`. [Browsers that support preload](http://caniuse.com/#search=preload) will fetch the font CSS file, but wait to actually apply it until it has loaded. Browsers that don't yet support preload get this functionality via Filament Group's [loadCSS](https://github.com/filamentgroup/loadCSS/) and [cssrelpreload polyfill](https://github.com/filamentgroup/loadCSS/blob/master/src/cssrelpreload.js).

I've also added [Font Face Observer](https://github.com/bramstein/fontfaceobserver) to improve the loading and handling of our web fonts. Currently, we make users wait to see any text on the page until after the web fonts have loaded. While this does avoid a Flash of Unstyled Text (FOUT), it decreases _perceived_ performance. Instead, we can embrace the FOUT and load the fallback fonts first, then let Font Face Observer watch for the Cloud.typography fonts to be loaded. When they're ready, it adds the `wdn-fonts-loaded` class to the `<html>` and the fallback font stack is replaced with the full font stack, avoiding a Flash of Invisible Text (FOIT). This improvement can/will be applied to the overall framework and is not reliant upon using the speedy template.

While the `rel="preload"` approach works well for the Cloud.typography fonts, I've yet to find an optimal solution for loading the all.css file. I've tried a number of different options (I feel like I've exhausted them all) and here are some observations:

- If I remove the framework all.css file and load only the web fonts using the above method, our PageSpeed scores shoot up to around 96/97 for mobile and desktop. 
- If I don't use any scripts (loadCSS, cssrelpreload, onloadCSS, load-deferred-css) and load the fonts.css and all.css using only `rel="preload"`, our PageSpeed scores again shoot up to around 96/97 for mobile and desktop. Of course, this isn't a viable option as preload isn't widely supported. However, I'm interested in the reasoning for this outcome.
- I also tried [Fetch Inject](https://github.com/jhabdas/fetch-inject) in lieu of loadCSS. While it brought our PS scores up into the high 90s again, its method of injecting the styles inline will break the relative URL paths for the icon font files (WOFF & WOFF2).
- PageSpeed Insights inexplicably shows two instances each for the all.css and fonts.css files. The page source and inspector show only one file each.
- The all.js file (moved to the bottom of the page, in the `<body>`) is shown as render-blocking when the all.css file is loaded via loadCSS. This led me to investigate the connection between these two files, the difference being the addition of the Font Face Observer. By removing it and  fontsloaded.js and inlining them instead, our PageSpeed score tops out at 100 for mobile and 89 for desktop. Now, get _this_. Google says our HTML… _is minified for mobile?!?!_ It very much is not, at least not _yet_. ¯\_(ツ)_/¯ Given the scores, I'd be inclined to go ahead and inline these font-handling scripts, but it occurred to me that doing so would break the web fonts (now keyed to the `'wdn-fonts-loaded'` class) for everyone not using the speedy template.
- After discussing with @mfairchild365, my hunch is that there is further optimization needed in the all.css and all.js files, but that will require additional investigation.
- I have my doubts about the reliability of using [sessionStorage](https://html.spec.whatwg.org/multipage/webstorage.html#dom-sessionstorage) for loading cached CSS on return visits. I imagine that in most cases it will work just fine, but I worry about the potential for broken styles if we make an update to the framework CSS and a visitor still has the key/value pair in their browser storage. I'm not even sure setting an expiration will overcome this.
> The lifetime of a browsing context can be unrelated to the lifetime of the actual user agent process itself, as the user agent can support resuming sessions after a restart.
- The critical CSS is for above-the-fold framework styles and doesn't account for above-the-fold content on individual pages using the framework. I'm curious how much these scores could drop on account of this.

**Mobile (96):** 
screen shot 2017-04-21 at 3 03 39 pm

**Desktop (88):** 
screen shot 2017-04-21 at 3 03 36 pm


mfairchild365:
Two things:

1) To address your concern about caching and session storage: as long as the `$DEP_VERSION$` is getting smudged in the build process correctly, I don't think there will be an issue. In the case where styles are marked as cached in the session storage, we are still loading the css via `https://unlcms.unl.edu/wdn/templates_4.1/css/all.css?dep=$DEP_VERSION$`. That URL will change when a new version is deployed, which will force cache-busting. It will result in a less than optimal page load for one request, but nothing should 'break'.
2) I think you want to merge this into the `speedy` branch instead of `develop`.

skoolbus39:
@mfairchild365 SMH. I thought I had chosen 'speedy' instead of 'develop.' I've updated the pull request.

PR #1066, opened 2017-05-08
Send focus to 'mobile' search input, avoid zoom

skoolbus39:
When a user presses the mobile search toggle, the dropdown search form widget opens, but focus remains on the toggle button. @mfairchild365 updated the JavaScript to send focus to the input. However, [when the input font-size is under 16px](http://stackoverflow.com/questions/11064237/prevent-iphone-from-zooming-form), iOS zooms to the input on focus. I've increased the 'mobile' search input size and changed related unit values from ems to rems to account for the updated font-size.


skoolbus39:
[Mobile search position: send focus to field](https://trello.com/c/ypgIDTai/75-mobile-search-position-send-focus-to-field)

PR #1069, opened 2017-05-11
Use double quotes for attribute selectors

skoolbus39:
- layouts/footer.less
- [states/idm.less](https://github.com/unl/wdntemplates/pull/1012#r74086184)
- [states/search.less](https://github.com/unl/wdntemplates/pull/1012#r74085726)
- states/share.less

### DIFF
--- a/wdn/templates_4.1/less/layouts/footer.less
+++ b/wdn/templates_4.1/less/layouts/footer.less
@@ -42,7 +42,7 @@
         }
     }
 
-    [class*='wdn-icon-'] {
+    [class*="wdn-icon-"] {
 	    position: relative;
 		padding-left: 1.777rem;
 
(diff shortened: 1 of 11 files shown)


wdn/templates_4.1/less/layouts/footer.less
```diff
@@ -55,7 +55,7 @@
         }
     }
 
-	span[class*='wdn-icon-'][aria-hidden='true']:before {
+	span[class*='wdn-icon-'][aria-hidden="true"]:before {
```
tlemburg: Do we want double quotes on wdn-icon- as well?

skoolbus39: Good call. I've applied them to `class`, `role`, and `type` attribute selectors, not just ARIA states.

PR #1071, opened 2017-05-18
Smudge speedy_body_scripts.html

skoolbus39:
Smudge file to replace '$DEP_VERSION' with version number.


skoolbus39:
[speedy template's dependents is not being smudged. This could lead to caching problems if not fixed.](https://trello.com/c/mq9rJTJl/87-speedy-template-s-dependents-is-not-being-smudged-this-could-lead-to-caching-problems-if-not-fixed)

PR #1073, opened 2017-05-23
Add delay to deleting Critical CSS when all is cached

tlemburg:
When doing a hard refresh on speedy template, there
is a flash of NO CSS. This occurs because the javascript
quickly throws in the deferred CSS to be loaded from cache,
then immediately deletes the critical CSS, which leaves no CSS
until the deferred CSS is loaded.

This change adds a 1 second delay before removing this CSS. This
should be enough time for almost all computers and phones to get
their stylesheet, parse, and render it to avoid this flash of no CSS.

I looked at simply not deleting the critical CSS, but that was a bigger
change that didn't need to be made at this time and without understanding
everything that is placed in each part, it's probably not a great
idea to tinker with it at this time.


mfairchild365:
This might help, but there is no guarantee that the the deferred CSS will be loaded after 1 second, especially on slower connections. Why not use a callback?

tlemburg:
I didn't want to use loadCSS here because that would put the stylesheet at a different spot than in the head tag: "The code above will insert a new CSS stylesheet link after the last stylesheet or script that it finds in the page". I just realized though that it actually does the same thing. I also kinda forgot that if we were hard-refreshing, you wouldn't be pulling the all.css from the cache at all. So the description above doesn't make much sense.

So I think the better plan then is to totally just forget about the sessionStorage thing, since it really doesn't do anything

wdn/templates_4.1/scripts/load-deferred-css.js
```diff
@@ -1,25 +1,11 @@
 (function() {
-
-    var WDNLoadDeferredStyles = function() {
-        var sheet = document.createElement('link');
-        sheet.rel = 'stylesheet';
-        sheet.href = 'https://unlcms.unl.edu/wdn/templates_4.1/css/all.css?dep=$DEP_VERSION$';
-        document.getElementsByTagName('head')[0].appendChild(sheet);
-    };
-
     var WDNRemoveCriticalStyles = function() {
```
mfairchild365: Now that this is only being used in one spot, it doesn't make much sense for it to be its own function.

PR #1075, opened 2017-05-30
Restore button cursor lost in normalize.css update

skoolbus39:
`cursor: pointer` was removed in a recent update to normalize.css. This commit restores the style for buttons and related form elements.

### DIFF
--- a/wdn/templates_4.1/less/layouts/forms.less
+++ b/wdn/templates_4.1/less/layouts/forms.less
@@ -1,5 +1,5 @@
 button,
-html [type="button"],
+[type="button"],
 [type="reset"],
 [type="submit"] {
     cursor: pointer;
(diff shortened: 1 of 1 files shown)


wdn/templates_4.1/less/layouts/forms.less
```diff
@@ -1,4 +1,11 @@
 button,
+html [type="button"],
```
mfairchild365: why the explicit scoping to `html`?

skoolbus39: I was restoring the removed styles directly from [normalize.css](https://github.com/necolas/normalize.css/blob/master/normalize.css#L278). However, the explicit scoping to `html` is to [prevent a WebKit audio/video control bug in Android 4](https://github.com/necolas/normalize.css/blob/master/normalize.css#L272-L273) caused by [-webkit-appearance: button](https://github.com/necolas/normalize.css/blob/master/normalize.css#L281). This fix is still part of normalize.css and still present in our framework, making the `html` unnecessary. I will remove.

PR #1084, opened 2017-10-18
Always show a focus outline for inputs

mfairchild365:
type=radio,checkbox,submit were not showing a focus outline.

### DIFF
--- a/wdn/templates_4.1/less/layouts/forms.less
+++ b/wdn/templates_4.1/less/layouts/forms.less
@@ -93,7 +93,7 @@ form {
         .sans-serif-font();
         font-size: .9353rem; // 12px
     }
-
+    
     input,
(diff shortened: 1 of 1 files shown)


mfairchild365:
@skoolbus39 can you review this? Also, do you know why we are using borders instead of outlines to show focus indicators?

skoolbus39:
I'm still not seeing a focus _outline_ for any fields in Chrome or Safari. 

The decision to use borders instead of outlines was not intentional on my part. This should be changed.

mfairchild365:
@skoolbus39 correct. This PR doesn't convert the focus borders to outlines, it just fixes a case where some types of inputs were not receiving any focus indication. A conversion from borders to outlines should be a different PR.

Issue #1062, opened 2017-04-20
Chrome prints responsive tables in collapsed format


ericras:
The different 'width' of the print view in different browsers is goofy.  Firefox prints the full table while Chrome prints tables responsively (if .wdn_responsive_table is applied).

I think in all cases the print version should render the full table and let the browser do the work of squeezing it as best as possible for print.  Collapsed responsive tables are a compromise for small screens but the print view more closely resembles a desktop view.

I'd lean towards wrapping .wdn_responsive_table in @media screen { }

PR #1086, opened 2017-11-27
Table styling updates

skoolbus39:
Improve color contrast for table header cells. At browsers widths under the ‘bp768’ breakpoint, the responsive table headers were white on a light background. At browser widths over the ‘bp768’ breakpoint, each responsive table would override the white with its own respective color and background-color. Now, the white table headers have been removed from the table body, and the contrast compliant color and background-color are applied at all browser widths, not just those above the ‘bp768’ breakpoint.

Borders are now present for table header cells in the table body for all tables, not just responsive tables.

The border-bottom has been removed from the abbreviation in the table header used to create responsive tables.

Coding style:
* New line for each selector.
* New lines in nested Less rules.

### DIFF
--- a/wdn/templates_4.1/less/layouts/tables.less
+++ b/wdn/templates_4.1/less/layouts/tables.less
@@ -60,8 +60,7 @@ table.wdn_responsive_table {
     }
 }
 
-
-@media @bp-under768w {
+@media screen and (max-width: 47.99em) {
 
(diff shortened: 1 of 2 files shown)


skoolbus39:
I figured while we're updating tables, we might as well wrap the responsive tables in screen-specific media queries, as suggested by @ericras.

Closes #1062

PR #1095, opened 2018-01-16
improve the skip link

mfairchild365:
uses the [skipto](https://github.com/paypal/skipto/) library to generate a more detailed skip navigation for sighted keyboard-only users. This could also help developers visualize the importance of correct heading semantics.


tlemburg:
Seems to be breaking. 

Uncaught TypeError: Cannot read property 'length' of undefined
    at Object.getAccessibleName (skipto.min.js:7)
    at Object.getLandMarks (skipto.min.js:7)
    at Object.init (skipto.min.js:7)
    at skipto.min.js:7
    at skipto.min.js:7

skoolbus39:
As we discussed, I'd like to eventually see it more closely align visually with 5.0, but that shouldn't stop us from including it now.

PR #1097, opened 2018-01-16
5.0 nav

tlemburg:
First pass at JS and functionality for 5.0 primary navigation. Feel free to make comments if I didn't use the framework correctly

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -318,14 +318,14 @@ module.exports = function (grunt) {
 				files: [templateScss + '/**/*.scss', templateJs + '/js-css/*.scss'],
 				tasks: ['sass']
 			},
-			js: {
-				files: [templateJs + '/**/*.js', '!' + templateCompileJs + '/**/*.js'],
-				tasks: ['js']
-			},
-			includes: {
-				files: [buildDir + '/**/*.html', templateScss + '/**/*.scss', templateJs + '/js-css/*.scss'],
-				tasks: ['includes']
-			}
+//			js: {
+//				files: [templateJs + '/**/*.js', '!' + templateCompileJs + '/**/*.js'],
+//				tasks: ['js']
+//			},
+//			includes: {
+//				files: [buildDir + '/**/*.html', templateScss + '/**/*.scss', templateJs + '/js-css/*.scss'],
+//				tasks: ['includes']
+//			}
 		}
 	});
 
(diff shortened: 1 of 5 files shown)


tlemburg:
Just remembered I need to do focus states

Gruntfile.js
```diff
@@ -322,10 +322,10 @@ module.exports = function (grunt) {
 				files: [templateJs + '/**/*.js', '!' + templateCompileJs + '/**/*.js'],
 				tasks: ['js']
 			},
-			includes: {
-				files: [buildDir + '/**/*.html', templateScss + '/**/*.scss', templateJs + '/js-css/*.scss'],
-				tasks: ['includes']
-			}
+//			includes: {
```
mfairchild365: Was this actually breaking something, or just doing nothing?

wdn/templates_5.0/includes/local/nav-primary.html
```diff
@@ -55,3 +55,7 @@
     </ul>
   </li>
 </ul>
+
+<div class="dcf-u-txt-center">
+  <button id="unl-primary-nav-close" class="dcf-c-btn dcf-c-btn--primary unl-c-btn-close-primary-nav" href="#" role="button">Close</a>
```
mfairchild365: The `role="button"` is redundant here, and it doesn't need the `href="#"`.

wdn/templates_5.0/js/main.js
```diff
@@ -7,3 +7,51 @@ requirejs.config({
 		}
 	}
 });
+
+require(['jquery'], function($) {
```
mfairchild365: This should probably live in a file named something like navigation.js instead of main.js.

wdn/templates_5.0/js/main.js
```diff
@@ -7,3 +7,51 @@ requirejs.config({
...
+	$(document).ready(function() {
+		var menu_shown_at_small = false;
+		var unl_primary_nav = $('#unl-primary-nav');
+
+		$('#dcf-menu-toggle').click(function (click) {
```
mfairchild365: A few things re-accessibility:
1) the button to toggle should have an `aria-expanded` attribute
2) the when the nav is opened, focus should be sent to the first link
3) when nav is closed, focus should be sent back to the button that opened it
4) closing on escape might also be helpful (especially for mobile nav when a bluetooth keyboard is being used)

You might be able to easily implement this with the dropdown widget JS I made for the IdM widget. https://github.com/mfairchild365/wdntemplates/blob/e8305700a3a3c5459f1297108a58c644819853bf/wdn/templates_5.0/js/dropdown-widget.js

tlemburg:
Reminder for myself: also need to change the "t" designations on things

PR #1102, opened 2018-01-26
Improve the structure of the js directory now that we are using babel

mfairchild365:
I think this should be reviewed+merged before other PRs, otherwise this will keep generating merge conflicts.

With the addition of babel transpiling, we were left with a somewhat messy `/js` directory. We had '/js/babelTranspile` and `/js/babelNoTranspile` directories, which were processed by grunt and spat back into the `/js` directory (combining the two folders into one flattened structure). In addition to those directories, we also had `/js/plugins` and `/js/require-css` directories. Additionally, the js files that were compiled with grunt were not being ignored by git.

This structure, while it works, could be very confusing to someone new to the project, as it is unclear where the source js files live and how they relate to the `/js` directory. In other words, it was hard to tell what was being compiled by grunt, and what was not.

This PR solve this by doing two things:

1. all js source files live in `/js-src`, outside of `/js`.
2. all files that should be transpiled via babel are given the `.babel.js` extension. All other files are simply copied to `/js` with grunt. This helps to flatten the `/js-src` directory so that its structure will more accurately reflect that of the `/js` directory.


Shaderpixel:
How are the files that are not being transpiled by babel get moved into the templateJS?

Shaderpixel:
Nvm saw the updated copy task. I like your suggestions.
Things need to be updated:
Gruntfile.js [lines 126-128](https://github.com/mfairchild365/wdntemplates/blob/71d260a41458dd901f28eed5af4ea4e589ae3410/Gruntfile.js#L126), remove the comments and the FileExclusionRegExp property since those folders are no longer in templateJS

Other than that it should be good. I will take another pass at it after you updated them.

mfairchild365:
@Shaderpixel good catch. I've removed that bit.

PR #1112, opened 2018-03-08
IdM widget styling

skoolbus39:
Because the upcoming usability tests and brand consultation will impact the layout and aesthetics of the header, the styling here is intentionally kept to a minimum. Additional styling will be added at a later date. The attributes have been reordered according to [http://codeguide.co/#html-attribute-order](http://codeguide.co/#html-attribute-order).

### DIFF
--- a/wdn/templates_5.0/js-src/idm.babel.js
+++ b/wdn/templates_5.0/js-src/idm.babel.js
@@ -267,7 +267,7 @@ define(['wdn', 'jquery', 'dropdown-widget', 'require'], function (WDN, $, DropDo
 
 			//Add the profile link
 			$navUL.append($('<li>').append($('<a>', {
-  			'class': 'dcf-u-inverse',
+				'class': 'dcf-u-inverse',
 				'href': this.getProfileURL()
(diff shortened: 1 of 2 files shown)


wdn/templates_5.0/js-src/idm.babel.js
```diff
@@ -255,24 +256,26 @@ define(['wdn', 'jquery', 'dropdown-widget', 'require'], function (WDN, $, DropDo
...
 			});
 
 			var $navUL = $('<ul>', {
-				'class': 'dcf-c-list-unstyled'
+				'class': 'dcf-o-list-bare dcf-u-mb0 dcf-u-sm2 unl-u-font-sans'
 			});
 
 			//Add the profile link
 			$navUL.append($('<li>').append($('<a>', {
+  			'class': 'dcf-u-inverse',
```
mfairchild365: this property should be indented.

skoolbus39: Fixed with f7c87c8

PR #1120, opened 2018-04-12
Add 4.1 deprecated styles (bands, grid) using the original LESS files

ericras:
A new CSS file will be created at css/deprecated/deprecated.css when grunt is run.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .settings
 .project
 .buildpath
+.idea
 /sandbox
 /build/compressed
 /build/vendor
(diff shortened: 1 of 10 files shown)


ericras:
I put the files in a /less/deprecated folder with the output as /css/deprecated/deprecated.css

I imagine in the future when there is deprecated SASS from early versions of DCF that there could also be a /scss/deprecated folder that appends its output to /css/deprecated/deprecated.css as well.

ericras:
Superceded by #1122

PR #1122, opened 2018-06-07
Add initial 4.1 deprecated styles to 5.0

skoolbus39:
Deprecated styles still need to be added for plugins/widgets and to handle infographics and icons.


ericras:
We discussed that ultimately the deprecated css should live in its own file separate from all.css as mentioned in the TODO in all.scss

PR #1137, opened 2018-08-10
Add clarification comments to gruntfile

Shaderpixel:
Added comments to clarify RequireJS config object and related variables.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -26,20 +26,20 @@ module.exports = function (grunt) {
 
 	// project layout variables (directories)
 	var mainDir = 'wdn',                                    // wdn
-		buildDir = 'build',                                   // build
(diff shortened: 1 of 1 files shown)


Gruntfile.js
```diff
@@ -384,7 +393,7 @@ module.exports = function (grunt) {
 		watch: {
 			sass: {
 				files: [templateScss + '/**/*.scss', templateJs + '/js-css/*.scss'],
-				tasks: ['scssGlobber','sass']
+				tasks: ['sassGlobber:all','sass']
```
mfairchild365: This doesn't appear to be a documentation change. What is the purpose of this change?

Shaderpixel: Just making the task scope be more specific and only run the 'all' task defined under scssGlobber instead of running the entire scssGlobber task (at this time there's only 'all' subtask, future there could be others).

PR #1141, opened 2018-08-31
Update 5.0 plugins, styles, webfonts

skoolbus39:
- Update node modules: DCF, Sass
- Initial work on mapping 4.1 widgets/plugins to 5.0
- Update form, color and other general theme styles
- Add Liberator webfonts and inline styles, scripts


ux-test.shtml
```diff
@@ -130,20 +130,20 @@
     <p>Authentic hoodie fap literally umami, Marfa Intelligentsia tousled sartorial keytar. Cliche dreamcatcher four loko bicycle rights, direct trade bespoke Neutra aesthetic quinoa fanny pack.</p>
 
     <h2>Vinyl bitters banjo, meggings forage Terry Richardson kogi pug beard aesthetic kitsch whatever selvage.</h2>
-    <p>Distillery XOXO Pinterest before they sold out single-origin coffee ugh, ethical small batch organic gentrify <abbr class="dcf-txt-sm" title="Video Home System">VHS</abbr>. Kale chips <a href="#">mixtape master</a> cleanse jean shorts Echo Park. Meggings Banksy umami, +1 you probably haven’t heard of them cliche gastropub keytar Neutra.</p>
+    <p>Distillery XOXO Pinterest before they sold out single-origin coffee ugh, ethical small batch organic gentrify <abbr class="dcf-txt-sm">VHS</abbr>. Kale chips <a href="#">mixtape master</a> cleanse jean shorts Echo Park. Meggings Banksy umami, +1 you probably haven’t heard of them cliche gastropub keytar Neutra.</p>
```
mfairchild365: So, the `title` attribute is actually a good thing with `<abbr>`. Without it, you might as well not use `<abbr>`. It is best to expand the abbreviation in the text on first use ("Video Home System (VHS)"), then use `<abbr>` with the `title` attribute in subsequent usage. This way, understanding the abbreviation is not solely dependant on the `title` attribute, but it might be a bonus in certain situations.

skoolbus39: That was my thinking, too, but I reconsidered after reading Aaron Gustafson's [A List Apart article](https://alistapart.com/article/conversational-semantics#section3) yesterday. He, too, recommends expanding the abbreviation on first use, but instead using a `<dfn>` element instead of an `<abbr>`. Thoughts?

PR #1142, opened 2018-08-31
Bring in dcf modules into 5.0

Shaderpixel:
- UMD-wrapped and transpiled modules can now be brought in from DCF


mfairchild365:
@Shaderpixel `npm ci` fails with the error "ci can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing."

I'm guessing this is due to the way the merge conflict was resolved earlier. Are you able to fix this?

Issue #1094, opened 2018-01-11
Search Results Layout


tlemburg:
Laying out search results for the 5.0 framework

ericras:
<img width="1372" alt="screen shot 2018-10-19 at 1 08 18 pm" src="https://user-images.githubusercontent.com/446357/47235912-14ae5e80-d3a0-11e8-8483-05f70290a1e6.png">

PR #1180, opened 2018-10-25
Fix search results

jsturek:
* Added styles to iframe to fill containing div with no border
* Added logic to remove search results on search close
* Updated setting siteHomepage.


ericras:
#1094

PR #1187, opened 2018-11-08
Remove small caps, display and script font JS/CSS

skoolbus39:
Remove optionally-loaded extended font packages—except extended Latin for Mercury Text.


jsturek:
Just an FYI I see Mercury references (that don't appear to be extended latin) in `wdn/templates_5.0/js/plugins/ui/css/jquery-ui-wdn.css`.  Other than that possible issue, looks good to me.

PR #1194, opened 2018-11-12
Update templates

jsturek:


jsturek:
Do we have a way to test .jsp files?

ericras:
> Do we have a way to test .jsp files?

No, idk who specifically is using them and they may very well just rip apart the cod eto fit it into their app anyway

wdn/templates_5.0/includes/global/head-2.html
```diff
@@ -2,13 +2,13 @@
...
 </style>
-<link id="core-css" rel="preload" href="/wdn/templates_5.0/css/core.css?dep=$DEP_VERSION$" as="style" onload="this.onload=null;this.rel='stylesheet';">
+<link id="core-css" rel="preload" href="https://unlcms.unl.edu//wdn/templates_5.0/css/core.css?dep=$DEP_VERSION$" as="style" onload="this.onload=null;this.rel='stylesheet';">
```
ericras: Have some double slashes here and below

jsturek: fixed

PR #1205, opened 2018-11-15
Fix body scroll bug

jsturek:
Fix body scroll bug with desktop nav closing without clicking close and resizing between mobile and desktop.

To test:
* In desktop verify that when nav is open and nav closes without clicking it closed, that the body scroll is not locked.
* Verify that moving between desktop and mobile with nav open that nav and body scroll behave correctly.
*  Verify that moving between mobile and desktop with nav open that nav and body scroll behave correctly.
*  Verify that moving between desktop and mobile with nav closed that nav and body scroll behave correctly.
*  Verify that moving between mobile and desktop with nav closed that nav and body scroll behave correctly.


skoolbus39:
We discovered another bug (#1215) effecting mega menu states, but it shouldn't hold up merging this fix.

PR #1212, opened 2018-11-20
use non-minified version of skipto.js in js-src

jsturek:
I noticed we had a minified version of skip to in our `js-src/plugins/skipto` directory.  I've been getting javascript errors related to skipto when trying to implement 5.0 in our apps on my local machine.  Not sure if this will fix those issues, but at least it will be structured similar to our other plugins.

You can test with `debug.shtml` if you use the local jsbody. `/wdn/templates_5.0/includes/global/js-body-local.html`


skoolbus39:
Looks good to me, too.

PR #1219, opened 2018-11-28
Flat pickr implementation

Shaderpixel:
- Added vendor copy tasks for vendor files from DCF
- Added example Flatpickr implementation to debug.html
- Replaced js-body.html in global include to js-body-local.html for local testing


jsturek:
Looks good, other than test failure.  I have the same issue on my PR.  Tests pass locally for me on both our branches.

PR #1242, opened 2018-12-14
shib.unl.edu URLs need a service parameter for when JS is disabled

ericras:


ericras:
Then merge it!

ericras:
The other reason for this is to not blow up webaudit when it crawls the link that 500s.

skoolbus39:
Maybe I will. Maybe I won't.

PR #1249, opened 2018-12-20
Add number and tel input types to input styles

natehilker:
Missing generic input styles for number and tel inputs.


skoolbus39:
Thanks, @natehilker

PR #1266, opened 2019-01-10
Add trailing slash when constructing siteHomepage

ericras:
The test that compares siteHomepage to www.unl.edu looks for a trailing slash and that's what existed before in 4.1 when siteHomepage was generated by the nav JS


ericras:
>  I found the same – a search for us at the top level doesn’t produce a link to us unless you click the ‘All of UNL’ filter, which some people aren’t aware of, and I think top level pages contain links to us.
>  
>  
> Who do I contact regarding the revised UNL web page? The search function on the new page doesn’t work very well. For example, if I type College of Arts and Sciences into the bar, I get a list of things that might be associated with the college (e.g., departments) but not a link to the college website. I have to go through a series of clicks to get to it. In some cases, it’s quicker to use google to find something within the UNL website than to use the search bar. I would say that the previous version was much more functional in this regard.

PR #1272, opened 2019-01-23
Adjust nav, search modal states

skoolbus39:
- Update ARIA labels for nav toggle buttons and modal states
- Update nav toggle button icons and toggle display to reflect button states
- Change search results <dialog> to <div>
- Update search results markup to fix scrolling bugs in iOS Safari
- Style modal states and transitions
- Update related critical CSS


ericras:
dialog-helper.babel.js can be removed right?

skoolbus39:
@ericras Eventually, yes, but it's still "in use" by DCF. I think it'd be safer to remove it there first, then remove remaining references to it in WDN.

PR #1280, opened 2019-02-08
Remove SkipTo plugin, replace with skip nav link

skoolbus39:
The SkipTo JavaScript plugin is generating duplicate IDs causing sites to fail HTML validity tests, thus bringing down WebAudit scores. This removes the plugin and replaces it with a simple skip nav link (what we had in version 4.1 of the framework).


ericras:
Closing for now after https://github.com/unl/wdntemplates/pull/1282 was merged.

cmdprompt:
This is still causing duplicate IDs to be generated from cloned source content that isn't supposed to have IDs in it.

cmdprompt:
I think this plugin is ultimately going to need to be removed, even after the update to how it generates IDs. Because after the initial nav is built, it doesn't keep up to date with DOM changes that occur in the document after its creation.

Consider DOM changes that would impact the skip to nav:
- could point to elements that no longer exist or are hidden in the current document state
- could reflect a nav structure that points to content in the wrong order due to DOM changes
- could contribute to invalid markup being generated by end user scripts because content is being cloned from markup that shouldn't have IDs in it

PR #1281, opened 2019-02-11
Fix WDN examples page

### LINKED ISSUE #1261: Needs fixing: /wdn/templates_5.0/examples/

skoolbus39:
While this fixes #1261, much of the documentation still needs to be updated and more examples are needed.


wdn/templates_5.0/examples/audioplayer.html
```diff
@@ -1,17 +1,17 @@
 <!DOCTYPE html>
 <html>
-<head>
+  <head>
     <meta http-equiv="content-type" content="text/html;charset=UTF-8" />
     <title>Audio Player Example</title>
-</head>
-<body data-version="4.1">
+  </head>
+  <body data-version="4.1">
```
jsturek: This should be `data-version="5.0"`.  Need to update all occurrences of in other files also.

skoolbus39: I was waiting until the corresponding documentation was updated to change the `data-version` for each example but, in retrospect, I'm not sure it makes much of a difference. I'll update the rest.

jsturek:
- [ ] IDM Example get JS error when clicking link: Uncaught TypeError: idm.displayNotice is not a function`

jsturek:
Still see some examples with `wdn` class references:
* events.html
* tabs.html
* typography.html

`ajax.html` still has bad `data-version` value.

Some of these examples (ajax.html and typography) are not used and could be deleted or added to examples navigation.

skoolbus39:
This isn't a complete update. This fixes the server-side includes to get it working in 5.0 before this month's framework release. Documentation itself still needs to be more thoroughly updated.

Where are you finding `ajax.html`? I don't see it in the [examples directory](https://github.com/unl/wdntemplates/tree/develop/wdn/templates_5.0/examples).

PR #1287, opened 2019-03-01
Hide menu button if nav contains no children

skoolbus39:
Check if primary/local navigation contains no children <`ul>`s (i.e., secondary links in the mega menu). If none are present, then remove the menu button at "desktop" widths. (It's still necessary at "mobile" widths to toggle the primary navigation.)

Thanks to @Shaderpixel for his help on this.


jsturek:
I noticed when there is no secondary nav, menu button flickers (appears and disappears).  We may want to consider hidden by default and then displaying if secondary nav exists.

jsturek:
- [ ] Need to verify works correctly on IE 11

ericras:
I tested and it looks good to me

PR #1299, opened 2019-04-15
Update footer markup

skoolbus39:
- Use `<small>` element for copyright.
- Change `.dcf-footer-stripe` to `.unl-footer-stripe`. The footer stripe is a theme style, not a DCF style.


ericras:
@skoolbus39 did you intend for the table mixin to be included in this?

skoolbus39:
Those are changes from DCF to the tmp.scss files. Now that we've added these files, it might take a PR or two to shake out the updates… I think.

ericras:
i guess that file just gets replaced by `grunt` anyway

PR #1308, opened 2019-04-26
limit decimal precision in SASS calculations

Shaderpixel:


skoolbus39:
Thanks for doing this, @Shaderpixel, but we need to limit the decimal precision in the JS plugin CSS as well. I'm going to close this and open a new PR.

skoolbus39:
Thanks, too, to @cmdprompt for the suggestion.

PR #1312, opened 2019-05-06
Use 'loading' attribute to lazy load images

skoolbus39:
- Update DCF.
- Use `loading` attribute instead of `dcf-lazy-img` class for lazy loading.


skoolbus39:
Close for now until the `loading` attribute is [stable in Chrome](https://addyosmani.com/blog/lazy-loading/).

PR #1320, opened 2019-06-10
Update DCF and Node modules

skoolbus39:
- Adds styles for `<tfoot>` and responsive & scrollable tables
- Adds portrait orientation (9×16, 3×4) aspect ratios
- Adds mixin/utility for XL max-width
- Adds responsive suffixes to aspect ratio utilities
- Adds media query for `prefers-reduced-motion`


jsturek:
Should there be an addition to `/wdn/templates_5.0/scss/core.tmp.scss` of the new line: `@import "../../../node_modules/dcf/assets/dist/scss/generic/_generic.reduced-motion";`

skoolbus39:
Yes. I thought that had already been added in a previous pull request, but it looks like it hadn't. I'll make note of it in the primary comment.

PR #1336, opened 2019-07-23
Isaac JS Option 1: Inline script tag type defer (Don't Merge)

jsturek:


ncolgrove:
FYI: I created a [codepen](https://codepen.io/ncolgrove/pen/gVMVqX?editors=1010) of the scenario to sandbox

Initially I wondered about audit errors and code sanitation problems from wysiwyg editors. But this solution does follow the [data block requirements of script tags](https://www.w3.org/TR/2018/WD-html53-20180206/semantics-scripting.html) and follow [MIME type requirements](https://www.w3.org/TR/2018/WD-html53-20180206/semantics-scripting.html). It appears there could be [case sensitivity issues with the selection query](https://caniuse.com/#feat=css-case-insensitive). I would suggest testing that further if you move forward. 

If a user incorrectly types ‘type="text/dcf"’ the script will never execute. No warning, error message, or indication of wrongdoing. Even debuggger, or console.log wouldn’t execute as a data block. For those not understanding what that flag does, I could see that being a potential source of confusion.

I don’t know where, but it feels like there could be a cross-site scripting or injection concern in here somewhere? We are asking the page to process a data block of hidden text as a new script. That's what JavaScript is, but it feels weird this time.

I would vote against due to potential silent fails.

jsturek:
Not implementing this option. Closing PR.

PR #1337, opened 2019-07-23
Isaac JS Option 2: Script tag movit defer (Don't Merge)

jsturek:


ncolgrove:
FYI: I created a [codepen](https://codepen.io/ncolgrove/pen/PMGzqr?editors=1010) of the scenario to sandbox.

This solution requires the deferredScripts object to be in the global scope AND be initialized before any other inline scripts. The pollution in the global scope creates clutter, and the early initialization creates code separation and a requirement of extra code in the head.

This solution seems too close to the window.dispatchEvent function to not use the native functionality, unless there is a reason otherwise.

jsturek:
Not implementing this option. Closing PR.

PR #1338, opened 2019-07-23
Inline script custom event defer

jsturek:


ncolgrove:
FYI: I created a [codepen](https://codepen.io/ncolgrove/pen/aemoZm?editors=1010) of the scenario to sandbox.

I originally assumed the window.dispatchEvent(event) would fire asynchronously. However, [dispatchEvent fires synchronously and native events fire asynchronously](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent#Notes) -- I learned something new. This may not be a concern since it looks like the kickoff script will be inside an async require callback? I forked the [codepen for an async test if interested](https://codepen.io/ncolgrove/pen/XvjrwR?editors=1010).

Is there a reason you have to declare the new Event object in the header? From what I understand [events don't need to be initialized before listeners](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Creating_and_triggering_events#Creating_and_dispatching_events_dynamically). 

I would probably vote for this option. It seems the most developer oriented and the “least hackish” option.

jsturek:
@ncolgrove, Good catch on create/dispatch. I was blindly following the moveit option pattern and overlooked.  I agree that this is now the cleanest.  The only drawback is IE support creating events but we have a polyfill for that so should be good.

PR #1350, opened 2019-09-10
Add DCF-based modals; adjust search & nav modals

skoolbus39:
- Add official DCF modals plugin/widget
- Adjust search and navigation modals HTML/CSS/JS


jsturek:
- [ ] Chat is hidden when opening site navigation

PR #1388, opened 2020-02-26
Add D3.js library

skoolbus39:
D3 will need to be configured and data manipulations will need to be created individually.


jsturek:
Probably should remove the `d3.min.js` file it serves no purpose since we are compressing the source file.

PR #1403, opened 2020-03-27
Add styles scoped for :visited inverse buttons

skoolbus39:
Fix colors for `:visited` inverse buttons


jsturek:
I'm still seeing the issue on non-inverse link buttons.
Screen Shot 2020-03-27 at 4 40 47 PM

PR #1404, opened 2020-04-02
Prepend banner to body if no skip nav

jsturek:


jsturek:
@ericras @skoolbus39 - Do we want to prepend banner to the body if skip nav is missing for some reason?  The current code will error out and the banner will not display if skip nav is missing.

PR #1407, opened 2020-04-03
Add Liberator web font

skoolbus39:
- Add Liberator Heavy Italic font files
- Add link, inline style and script for Liberator


jsturek:
@skoolbus39 Verified font loads, but not data:application like the others (viewing in inspector Network).  Just an FYI, otherwise looks good.

PR #1409, opened 2020-04-09
Make banner xhr request async

jsturek:


wdn/templates_5.1/includes/global/header-global-1.html
```diff
@@ -3,7 +3,6 @@
     function displayWDNNoticeBanner() {
 	    var bannerEnabled = true;
 	    if (bannerEnabled) {
-		    var xhr = new XMLHttpRequest();
 		    var messageContent = false;
 		    var messageSessionKey = 'wndNoticeMessage';
```
ericras, line 7: didn't notice this before - 'wnd'

jsturek: You don't like the key name?

PR #1426, opened 2020-06-09
Visit / Apply / Give CTAs

skoolbus39:
Add Visit / Apply / Give CTAs to local nav ("mobile") and global header ("desktop").


jsturek:
Need to create a templates_5.2 directory and make changes there and undo changes to templates_5.1 directory

PR #1448, opened 2020-08-11
Temporarily disable dark mode

skoolbus39:
Temporarily disable dark mode to allow time for more testing.


jsturek:
https://web.dev/prefers-color-scheme/

jsturek:
This is fine but how are developers suppose to test their sites for dark mode? Should we try a way to have it available, but disabled with a with a way to enable?  We can do this for now but probably want to have a better solution available soon.

ericras:
https://github.com/unl/wdntemplates/tree/5.2-dark-mode-preview branch created for dark mode testing.

PR #1462, opened 2020-09-09
Add scroll animations to WDN theme

skoolbus39:
- Update DCF, Node modules
- Update animation, transition variables
- Update variables used for transitions
- Add scroll animations to WDN theme


wdn/templates_5.2/js-src/scroll-animations.babel.js
```diff
@@ -0,0 +1,10 @@
+require (['dcf-scrollAnimation'], function() {
```
jsturek, line 1: Make consistent with new plugins:
`require (['dcf-scrollAnimation'], () => {`

jsturek:
@skoolbus39: Make that small change and you can merge it.

Issue #1473, opened 2020-10-16
QA Test link not functioning properly in Chrome


kmcguffey2:
QA Test link taking user to scan of site's homepage instead of current page scan in Chrome.

jsturek:
Issue resolved with: https://github.com/unl/wdntemplates/pull/1474.  Should be available with Nov WDN release.

PR #1474, opened 2020-10-16
Set URL to test for QA Test link

jsturek:
Fix for issue: https://github.com/unl/wdntemplates/issues/1473

WebAudit relies on PHP `$_SERVER['HTTP_REFERER']` to determine which page the QA Test is for unless it's passed in via the query string in parameter `url`.  The PHP interpreted value varies between browsers since the browser set the value and some browsers (chrome) omit part of the page path.

This solution sets the `url` query  string param via javascript with the pages location info.


wdn/templates_5.2/includes/global/footer-global-2.html
```diff
@@ -42,3 +42,15 @@ <h2 id="dcf-footer-group-4-heading">Policies &amp; Reports</h2>
...
+  var qaTestLink = document.getElementById('qa-test');
+  if (qaTestLink) {
+    var pathname = document.location.pathname;
+    // webaudit expects pages to end with a slash, so add one if missing
+    if (!pathname.match(/\..*$/) && !pathname.match(/\/$/)) {
+      pathname += '/';
+    }
+    qaTestLink.search = '?url=' + encodeURI(document.location.origin + pathname);
+  }
+</script>
```
skoolbus39, line 56: No newline at end of file

skoolbus39 (changes requested):

wdn/templates_5.2/includes/global/footer-global-2.html
jsturek: @skoolbus39, added new line

skoolbus39 (approved):

PR #1475, opened 2020-10-19
Remove :focus styles for mouse users. Preserve browser defaults for keyboard users.

skoolbus39:
https://css-tricks.com/the-focus-visible-trick/


jsturek:
I'm still seeing the focus with mouse events.

skoolbus39:
Which browser are you using? https://caniuse.com/css-focus-visible

jsturek (approved):
Looks good

PR #1477, opened 2020-11-12
Add dark mode toggle bookmarklet CSS

skoolbus39:
Thanks, @ncolgrove and @beckyaiken, for your work on this!


jsturek:
I guess I'm fine with this, but I don't know what this really gets us since it just a static css file which will have to be manually updated if any of these values change.  I was expecting it to be part of the build process and pull css values from a centralized source.

skoolbus39:
@jsturek That could be phase two. By including it in the framework, I can update those values manually (for now) any time they are changed in the main CSS – instead of having Nic have to update them on his end.

PR #1481, opened 2020-11-18
Remove space below footer caused by tracking pixel

### LINKED ISSUE #1480: Fix extra vertical space caused by tracking pixel
Tracking pixel causes additional vertical space below the footer.

skoolbus39:
Fixes #1480

### DIFF
--- a/wdn/templates_5.2/scss/elements/_elements.images.scss
+++ b/wdn/templates_5.2/scss/elements/_elements.images.scss
@@ -3,6 +3,14 @@
 ////////////////////////////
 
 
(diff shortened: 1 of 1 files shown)


ericras (approved):

ericras:
I don't think we have a guarantee that the tracking pixel will be the last element on the page.  Other javascript could possibly also try to insert something at the end of the page.  Is there some way to more specifically target it? Like:

body > img[src$="trkn.us"]

ericras (approved):

PR #1486, opened 2020-12-04
Modal Video Open/Close Events

jsturek:
Use debug-clean.shtml to view/test sample video modals.  Note: These examples will be removed before merging. Similar examples can be added to debug.shtml if needed.


jsturek:
No you shouldn't.  I checked and got the error message and then it went away.  Maybe the video was updated? Anyway it works again for me.

jsturek:
You shouldn't. It appears it does not like the video.load() call. I'll look into more. Thanks

jsturek:
@skoolbus39, The media load issue has been resolved.

skoolbus39 (approved):

PR #1502, opened 2021-04-09
Add additional lengths (height, width, margin, padding)

skoolbus39:
- Add variables to accommodate new additional lengths for height, width, margin and padding
- Update DCF


jsturek (approved):
Looks good

PR #1503, opened 2021-04-13
Remove object-fit polyfill

skoolbus39:
Polyfill was causing font display issues if text was contained within elements that used object-fit. Object-fit is now supported by over 97% of browsers – and we no longer support Internet Explorer.


jsturek (approved):
Looks good

PR #1504, opened 2021-05-05
Underline footer links that aren’t list items

skoolbus39:
Footer paragraphs need to distinguish links with underlines.


jsturek (approved):
Looks good

PR #1510, opened 2021-05-12
Fix caption color in slideshows

skoolbus39:
Caption contains default `figcaption` color and fails color contrast compliance against the dark slideshow caption background. Thanks @beckyaiken for spotting this!


jsturek (approved):
Looks good

PR #1511, opened 2021-05-12
Implement DCF Card as Link

jsturek:


skoolbus39 (approved):
I will follow up with a separate pull request to add styling for the card-as-link.

PR #1518, opened 2021-06-21
NPM Update

jsturek:
Update NPM packages to latest versions.  Update now requires use of node v14.17.0 (was using v8.17.0).
The use of Babel 7 (from Babel 6) is one the major changes.

This update now also uses DCF 3.0 which has a breaking change on how the javascript behaves (the reason for the version bump).  The DCF 3.0 javascript files are now ES6 Modules.  Grunt is updated to convert those modules to ADM modules and the ADM usage of those files has been updated.

Due to major package changes carefully check that all CSS and JS files are rendering and behaving as desired.

@ericras, please check that changes will not impact WDN deployment process and if further updates are necessary to support deployment.


wdn/templates_5.3/js-src/datepickers.babel.js
```diff
@@ -1,7 +1,7 @@
-require(['dcf-utility', 'dcf-datepicker'], () => {
+require(['dcf-datepicker'], (datePickerModule) => {
   //Date Pickers
   const datePickers = document.querySelectorAll('.dcf-datepicker');
   datePickers.forEach(function (dp) {
-    new DCFDatepicker(dp);
+    new datePickerModule.DCFDatepicker(dp);
   });
 });
```
skoolbus39, line 7: No newline at end of file

skoolbus39 (changes requested):

skoolbus39 (approved):

PR #1521, opened 2021-07-02
Update footer colors & background-color

skoolbus39:
- Update footer colors & background-color
- Remove color utility class (handled by footer CSS)


jsturek (approved):
Looks good.

PR #1526, opened 2021-07-16
Add GSAP to plugins

skoolbus39:
Can be tested by adding the test `rotate-cw` class to an element, and then adding the following markup to the debug page:
```html
<script>
  window.addEventListener('inlineJSReady', function() {
    require(['plugins/gsap/gsap.min'], function() {
      gsap.to('.rotate-cw', {rotation: 360, duration: 90, ease: 'none', repeat: -1, transformOrigin: '50% 50%'});
    });
  });
</script>
```
The element with the `rotate-cw` class should slowly rotate.


jsturek (approved):
Looks good.

PR #1528, opened 2021-08-03
Hide icon when no time is set for event

skoolbus39:
Convert clock icon to variable and set to not display for all-day events or when no time is entered.


jsturek (approved):
Looks good

PR #1530, opened 2021-08-09
Update incident reporting link in footer from TIPS to UNL Report

skoolbus39:
- Update link from TIPS to UNL Report
- Replace TIPS logo with UNL Report lockup


jsturek (approved):
Looks good

PR #1535, opened 2021-08-27
Add utility for dots background images

skoolbus39:
- Apply `background-size` to all variations of dots `background-image` utility classes
- Add missing `!important` to `background-image` ~~utilites~~ mixins


jsturek (approved):
Looks good

PR #1541, opened 2021-09-14
(Temporarily?) remove datepicker date relative positioning & z-index

skoolbus39:
`postition: relative` now causes the `box-shadow` to effectively disappear in Chrome.


jsturek (approved):
Looks good

jsturek:
Probably should update DCF starter as well.  Can wait until tomorrow if higher priority items remain for today.

PR #1560, opened 2021-11-10
Fix heading color in notices

skoolbus39:
Currently, the default heading color is changed to cream (`#fefdfa`) for only the `.dcf-notice-heading`. This pull request applies the color change to all headings inside the notice – including those nested inside `.dcf-notice-message`.


jsturek (approved):
Looks good

PR #1562, opened 2021-11-30
Add `sizes="any”` to favicon.ico

skoolbus39:
Prevent browsers from double downloading both SVG and ICO icons: https://css-tricks.com/favicons-how-to-make-sure-browsers-only-download-the-svg-version/


jsturek (approved):
Looks good

PR #1573, opened 2022-02-08
Update border variables; update variable values in dark mode bookmarklet

skoolbus39:
- Update border variables – `var(--b)` is the same as `var(--b-light-gray)`
- Update variables in dark mode bookmarklet


jsturek (approved):
Looks good

PR #1580, opened 2022-02-21
Add focusable=“false” to slideshow SVG icons

skoolbus39:
- Update Node modules
- Add `focusable=“false”` to slideshow SVG icons


jsturek (approved):
Looks good

PR #1587, opened 2022-04-12
Update GTM script/noscript locations 

skoolbus39:
- Remove GTM `<script>` from `head-2-local.html`
- Move GTM `<noscript>` from `skipnav.html` to `js-body.html`


ericras:
This means that:
 - the "local" template will not have GTM, and
 -  the "fixed" (with unlcms.unl.edu references to the dependecnices) template will continue to have GTM code.

ericras (approved):

PR #1607, opened 2022-08-22
moved compressed js to files

tommyneu:
looks like i missed moving the compressed js to the head html files after fixing a typo


skoolbus39:
👍

PR #1609, opened 2022-08-30
Update print styles

skoolbus39:


ericras:
What about tabs? Should every tab section be shown and printed?

ericras (approved):

PR #1627, opened 2022-11-10
Fix editable region comments

skoolbus39:
Change `<!-- Instance` to `<!-- Template`


ericras:
The other template files have a mix of Instance and Template.  I'm not understanding what the difference is reading https://helpx.adobe.com/dreamweaver/using/recognizing-templates-template-based-documents.html and which should be used where.

Whatever the case all of our template files should be consistent.  Was this one out of line with the others?  I'll admit to not having looked yet myself at that.

skoolbus39:
@ericras This was [fixed](https://github.com/unl/wdntemplates/commit/cf0a670a0fd8db9e20a169941c55c838d7c2ba10) once before. I'm not sure where the `<!-- Instance` got reintroduced, but I _think_ `<!-- Template` should be used exclusively in the Template files. In the time you were writing your comment, I had also discovered the errors in the additional files and updated those as well.

PR #1638, opened 2023-01-26
Increased the search time from 1000ms to 2500

tommyneu:
Due to issues with google search we will try increase search typing auto search


tommyneu:
We will instead be getting rid of the auto submit #1640

PR #1640, opened 2023-02-07
Commented out the search auto submit

tommyneu:
Due to issues with the google custom search engine we are removing the auto submitting search results


skoolbus39:
Looks good to me 👍

PR #1689, opened 2023-07-14
Updated for new version of search

tommyneu:
Search will be updating but almost everything should be the same. The route to the embed version will be different, and the postMessage data for new searching will be different


skoolbus39 (approved):

tommyneu:
I updated Search 2 to redirect if it finds embed as a query param and it will work with the old version of the postMessage. So people who have not updated the framework will still have a working version.

PR #1690, opened 2023-07-17
Fix issues with initialize not running on certain domains

tommyneu:
Sometimes google will load its own gtag.js which causes IDM not to be initialized. This now checks a different script to check if it has been initialized already


skoolbus39 (approved):

ericras (approved):

ericras:
Background: it seems that GTM-WKHNC39 is set to load G-XYGRJGQFZK (at least sometimes on some domains, unclear) so this causes a return before idm.initialize executes.

PR #1814, opened 2025-07-07
1812 migrate alert component to wdn6.0 version 2

### LINKED ISSUE #1757: Alert close icon is shown after refresh
When you close the alert and refresh the page the close icon is shown instead of warning icon

### LINKED ISSUE #1812: Migrate alert component to wdn framework 6

sarongebre:
Migrated alert component to wdn framework 6. Replaced all jQuery with plain JavaScript. Fixed alert close icon showing after refresh. 

Version 2 contains: Saving alert data to a cookie to reduce the number of times the alert JavaScript needs to be loaded. Two new cookie names were created to prevent conflicts with existing WDN5 alert cookies, as some sites will remain on WDN5 for a while before migrating. 

Closes https://github.com/unl/wdntemplates/issues/1812
Closes https://github.com/unl/wdntemplates/issues/1757


wdn/templates_6.0/js-src/components/unl-alert.js
```diff
@@ -42,6 +39,61 @@ export default class UNLAlert {
...
+    // Sets a cookie to confirm that the 'unlalerts' JS was successfully loaded and data is received. 
+    #setAlertDataReceivedCookie(unlAlertsDataAlert) {
+        setCookie(
+            this.alertDataReceivedCookieName,
+            encodeURIComponent(JSON.stringify(unlAlertsDataAlert)),
+            this.alertDataReceivedCookieMaxAge,
+        );
+    }
+
+    async #checkIfFetchCallNeeded() {
```
tommyneu, line 63: If I have two tabs open and if there is no data in unlcap.js then both tabs will be fetching the data. If there is data in unlcap.js then it preforms as intended. This is leading to unneeded requests for unlcap.js. If there is no data in unlcap.js then the cookie needs to still be set so that these unneeded requests are not being sent

tommyneu (changes requested):

tommyneu (changes requested):
Making sure to write this down, the ID `unlalert` is being used in multiple elements

PR #1844, opened 2025-09-05
Fix debug-clean

tommyneu:
Debug clean was still using the old 5.3 template with invalid virtual includes


debug-clean.shtml
```diff
@@ -133,7 +119,7 @@
     <!-- InstanceBeginEditable name="maincontentarea" -->
     <p>Impress your audience with awesome content!</p>
     <!-- InstanceEndEditable -->
-   </div>
+    </div>
```
skoolbus39, line 122: Probably don't need a space here.

skoolbus39 (changes requested):

PR #1846, opened 2025-09-05
Remove `dcf-d-block` class from framework images

skoolbus39:
Images are already set to `display: block` by default in 6.0.


skoolbus39:
This PR pairs with https://github.com/digitalcampusframework/dcf/pull/558.

tommyneu (approved):

PR #1903, opened 2026-04-03
Add `dcf-d-none@print` to nav and footer regions

skoolbus39:
Pairs with https://github.com/digitalcampusframework/dcf/pull/591


tommyneu (changes requested):
- Notices look broken when printing, their icon end up being 100% width
- The off site link icon is conflicting with the print styles to have the link's href in the `:after`

tommyneu (approved):

